PR #72, opened 2014-07-18
Add instructions to bootstrap with a notification email

riking:
This solves the problem of getting started with Amazon SES email.


SamSaffron:
seems fine

riking:
Uh, actually, it isn't fine anymore, because we have a way to set this in the env.

PR #89, opened 2014-08-29
BUGFIX: if one of runit's 1.d scripts fails, abort the startup

Supermathie:


SamSaffron:
this seems fine to me

PR #99, opened 2014-10-15
Fix web tls.

wmark:


SamSaffron:
@wmark I want to merge this in but it needs to be rebased, I cant easily do this merge manually

wmark:
I've resumed work on this.

Session ticket rotation using ```nginx -s reload``` is not the way to go here because it has the side-effect of purging buffers which we still need. For example, the open-files and metadata cache.

wmark:
Abandoned.

This is why:

* I believe that curve~ and cipher suite settings don't belong in *nginx*. They depend on the TLS implementation, and should be set there.
* Any other compiled-in settings are best made by the release engineer of a distribution. The customer should find an minimal configuration file which is meant to contain only his/her specific differences. (Move default settings into *nginx*.)
* I've switched to *boringssl*, and have modified it.
* … which results in *nginx* and *Ruby* (git, curl, …) looking for different implementations at the same place
* … which is one of the prime anecdotes for creating (semi-)static binaries (compiling-in dependencies).

You can find the resulting *nginx* here. Some modifications ('auto' cpu-affinity, ticket rotation, TLS dynamic buffer(s) with caching, cert- and curve negotiation) have been open-sourced, everything else is trivial:
https://s.blitznote.com/debs/nginx-boringssl-1.9.4-0

```bash
add-apt-repository "deb https://s.blitznote.com/debs/ubuntu/amd64/ all/"
printf 'Package: *\nPin: origin "s.blitznote.com"\nPin-Priority: 510\n' \
  > /etc/apt/preferences.d/prefer-blitznote
# the packages are not signed with gpg, but with signify
```

I don't expect this to be integrated into Discourse. Therefore I abandon this PR.

PR #102, opened 2014-10-22
apt-get update before install to avoid launch failure on fresh image

lmsurprenant:
required if the package signature owner's public-key is not in the keyring (http://superuser.com/a/164580/35073).  This was preventing me from running './launch bootstrap app' with success on a fresh clone.


SamSaffron:
pushed whois into base image ... we should be good now

PR #103, opened 2014-11-03
sample for disabling starttls (local smtp, etc)

YesThatAllen:
Adding the sample here can prevent people from having to hunt this key down


SamSaffron:
completed via https://github.com/discourse/discourse_docker/commit/b1d030e5cdbdbe611a13c9038d3a1e4548183647

PR #105, opened 2014-11-09
Add --docker-args option to the launcher script

samdroid-apps:
Eg:

```
./launcher start app --docker-args "-m 4g --cpuset='6-7'"
```

This will limit discourse 4 gigs of RAM and cpu 6 and 7 only.

### DIFF
--- a/launcher
+++ b/launcher
@@ -53,6 +53,7 @@ usage () {
   echo
   echo "Options:"
   echo "    --skip-prereqs   Don't check prerequisites"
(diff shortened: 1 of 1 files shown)


SAMdroid-apps:
@SamSaffron, sorry about the late reply, I think it must of got lost in the mail!

I'm pretty sure it does, but I can't really test on my production instance :)
I can try to setup an instance on my laptop, but that will take some time.

SamSaffron:
ok, once you confirm I will merge it in.

PR #107, opened 2014-11-09
Fix package authentication in the app.yml files/samples.

Moter8:
http://i.imgur.com/AgPxHuG.png

`apt-get` returns an authentication error when the repos haven't been updated.


SamSaffron:
fixed via https://github.com/discourse/discourse_docker/commit/ce2f683ef7826b8aab22df9f6db0998d9eadaaed

PR #117, opened 2014-12-11
setting default gravatar url

fantasticfears:
I don't know whether pup overrides the env in order. It's required to make this work properly. Can you confirm this behaviour?


SamSaffron:
pups does override env in order, but we need to work out the parent PR on Discourse first.

PR #118, opened 2014-12-13
Add quotes around values in the generated discourse.conf file

malexw:
Previously, if a config value contained a # character, the hash and all the
following values would be treated as a comment. Wrapping all the values with
quotes will allow for hashes, which could be useful in the case of an SMTP
password, for example.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -31,7 +31,7 @@ run:
         conf=/var/www/discourse/config/discourse.conf
 
         # find DISCOURSE_ env vars, strip the leader, lowercase the key
-        /usr/local/bin/ruby -e 'ENV.each{|k,v| puts "#{$1.downcase} = #{v}" if k =~ /^DISCOURSE_(.*)/}' > $conf
+        /usr/local/bin/ruby -e 'ENV.each{|k,v| puts "#{$1.downcase} = \"#{v}\"" if k =~ /^DISCOURSE_(.*)/}' > $conf
 
   - file:
      path: /etc/service/unicorn/run


SamSaffron:
I am terrified of mucking with this code so it needs to be bullet proof, meaning you need to also handle bash escaping of newlines and `"`

lidel:
This PR is a fix for a bug discussed in https://meta.discourse.org/t/bug-smtp-password-field-does-not-escape-comment-sign-hash/23344

Wrapping config values with  `"` or `'` is already supported by Discourse. 
You can check [global_setting.rb#L69](https://github.com/discourse/discourse/blob/ff063f4f9b866f18be95a4cb173d2fcea2e046b8/app/models/global_setting.rb#L69) where config file is parsed.

Should be safe to merge: values that land in `discourse.conf` do not have newlines and `"` was already recognised by parser in `global_setting.rb` as a wrapping character (so there is no risk of breaking people's instances).

I tested it on my local instance and worked fine: bug is gone, SMTP password with `#` worked fine.

coding-horror:
I guess it is unlikely we are going to merge this at this point, so I will add a warning about # in passwords in the comments

lidel:
I agree. Just for the record, if someone is looking for a safe workaround:
https://meta.discourse.org/t/bug-smtp-password-field-does-not-escape-comment-sign-hash/23344/16?u=lidel

coding-horror:
I did this 

https://github.com/discourse/discourse_docker/commit/7439768eb05dcdb101677e913015dff811897093

PR #119, opened 2015-01-01
Set default encoding UTF8, not ASCII

audreyt:
This resolves the CJK indexing issues.

c.f.: https://meta.discourse.org/t/chinese-search-issues/13287/56


SamSaffron:
I can't accept this for a couple of reasons

1. why is it only on the 9.2 template? 9.3 is the template we really care about
2. this needs to be more specific, only change encoding if it matches a known bad one.

SamSaffron:
can you please reopen a corrected PR ?

PR #126, opened 2015-02-09
Add prompt when running on device mapper

riking:
This should give the user a fair chance to bail out when they thought that they had AUFS running but they don't and want to fix it.


LuaTenshi:
Maybe something like what is mentioned here can be added to this? http://unix.stackexchange.com/questions/26676/how-to-check-if-a-shell-is-login-interactive-batch

or maybe... http://bytefreaks.net/gnulinux/bash/bash-script-how-to-check-if-it-run-by-a-user-in-terminal-or-not

SamSaffron:
I will accept this, but lack of TTY check means that this will cause hanging if docker is not installed right.

PR #127, opened 2015-02-16
Update launcher

kewubenduben:
just a typo


SamSaffron:
thanks, we need to get rid of these cids, they are just trouble, especially the standard cid files

PR #130, opened 2015-03-10
* Set default encoding to UTF8, not SQL_ASCII

audreyt:
This is a cleaned-up patch as suggested by @SamSaffron in #119.


SamSaffron:
sure this seems fine ... thanks!

zdk:
:+1:

PR #131, opened 2015-03-10
Change web_only sample database username variable to what discourse is expecting

lfielke:
The sample `web_only.yml` container configuration file has a commented out example to change the Postgres username, however it is not the correct variable name as of the current version of discourse. This means that uncommenting it and changing the value has no effect when building the container.

The Discourse config parameter that changes the database username is `db_username` as per https://github.com/discourse/discourse/blob/ff063f4f9b866f18be95a4cb173d2fcea2e046b8/config/discourse_defaults.conf


tgxworld:
`db_name` is used in [`discourse_defaults.conf`](https://github.com/discourse/discourse/blob/ff063f4f9b866f18be95a4cb173d2fcea2e046b8/config/discourse_defaults.conf#L34).  I don't think `DISCOURSE_DB_NAME` was a mistake. :smile_cat:

tgxworld:
Opps ignore my previous comment. I read the variably name wrongly. :sweat_smile:

SamSaffron:
yeah, looks right to me .. .thanks https://github.com/discourse/discourse/blob/master/config/discourse_defaults.conf#L37

PR #141, opened 2015-04-03
Added a sample template web_redis.yml

nileshtrivedi:
As per the discussion here: https://meta.discourse.org/t/docker-setup-for-app-and-redis/27110/2 , it seems many people want this and have already been doing it themselves.


SamSaffron:
Sorry, I can not support this its just an odd odd grouping that needs to be manually maintained by whoever needs it. I dont want to maintain another sample here that is going to invariably go out of sync with the rest of the samples.

PR #147, opened 2015-04-21
templates/web.sssl.template.yml: Disable spdy header compression

rata:
As explained here:

https://github.com/18F/tls-standards/issues/24

We shouldn't use spdy header compression with spdy 1.3 (the one packaged with nginx).So, just disable it.


SamSaffron:
thanks!

PR #148, opened 2015-04-27
Revert templates/web.sssl.template.yml: Disable spdy header compression

rata:
This reverts commit c0ed494c3acf1acf629379f33f413d7660183855. It was late and I
misread the info here:

	https://github.com/18F/tls-standards/issues/24

The first comment clearly states that the server *can* compress spdy header. And
modern browsers disable compression on responses (or chrome, maybe smarter
tricks in newer versions). In fact, Chrome 21 and Firefox 15 switched off SPDY
header compression[1] and the requirements page on discourse:

	https://github.com/discourse/discourse#requirements

are newer browsers than that. So it seems safe to enable spdy header
compression on the server.

[1]: https://www.imperialviolet.org/2012/09/21/crime.html

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -26,8 +26,10 @@ run:
        ssl_session_tickets off;
        ssl_session_cache shared:SSL:1m;
 
-       # disable SPDY header compression (flawed in spdy < 4)
-       spdy_headers_comp 0;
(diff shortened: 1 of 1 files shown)


SamSaffron:
thanks

rata:
thank you!

PR #151, opened 2015-04-30
templates/web.sssl.template.yml: Improve cipher list

rata:
This changes the cipher list for a more readable one, and avoids using weaks
ciphers. As a side effect of this, the grade on ssllabs is now A+ instead of A.


rata:
Ping ? :)

Supermathie:
I'd recommend sticking with the [Intermediate](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29) compatibility list as a sensible default. The [Modern](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) list cuts out a large swath of potential audience and likely will only lead to frustration.

rata:
@Supermathie totally agree on that. But this change has nothing to do with it, if I'm not missing something.

The thing is that using that cipher list, without a dh_param file > 1024, causes the use of weak ciphers (marked weak by ssllabs). Just keeping the big list and adding the "!DHE:!EDH" thing at the end, I *think* it should fix it (didn't test it). Doing it the way the patch does, (using HIGH) fixes it too.

Adding a dh_param file I think is not as straight forward as any of these patches, as the file probably needs to be created when building/rebuilding the container. And if we keep that list as is, we are using these ciphers with weak params.

*I* think is something worth fixing/improving and adding the !DHE:!EDH thing should do the trick without breaking browsers. But we can also look at dh_param as another way to fix this, if we want to fix it.

I don't see any other ways to fix it thant these, but I might be missing something

rata:
Ohh, @SamSaffron not sure you will get a notification. So, just in case, see the comment I've just added here: https://github.com/rata/discourse_docker/commit/3cf5d461eb59006171cf5fb76708ff7931c0d73c

It's rather important to fix this issue now, as an attack that the default discourse install is vulnerable has been announced

rata:
The alternative of using dh_param was commited here: https://github.com/rata/discourse_docker/commit/421b9abef0fec6d495d009a82aa1311c43203cf3

So closing this.

Issue #149, opened 2015-04-28
Unattended upgrades not working ?


rata:
Hi,

I've installed discourse the docker way and did the unattended upgrades config as recommended in the guide. But there were updates available and after a few days there, there weren't upgraded.

Is cron running fine for unattended upgrades on the container ?Does it work for you ?

SamSaffron:
are you talking about updates inside the container, that is not implemented
and does not really fit well in a container world.

rata:
Yes, I'm taking about the recommendation here: https://github.com/discourse/discourse/blob/master/docs/INSTALL-cloud.md#post-install-maintenance

As you ship a "full" ubuntu LTS image, packages there need security upgrades. I'm doing them in some other way, but wanted to let you know that it seems the recommendation to keep them up to date is not working.

SamSaffron:
the recommendation there is about the host. 

I am torn on doing container upgrades in that way. In general we ship new images if any of the apps in our container need an upgrade.

rata:
Oh, ok. I'm interested in upgrading the container, not the host. I already manage that. Sorry I misunderstood :)

SamSaffron:
closing, assume this is resolved

PR #152, opened 2015-04-30
Improve documentation of ssh launcher command

rata:
The access via ssh is only granted if the user building (or rebuilding) has a
ssh key. So, make that clear in the "documentation" for that command.

### DIFF
--- a/launcher
+++ b/launcher
@@ -44,7 +44,7 @@ usage () {
   echo "    restart:    Restart a container"
   echo "    destroy:    Stop and remove a container"
   echo "    enter:      Use nsenter to enter a container"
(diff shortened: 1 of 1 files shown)


rata:
Ping? :)

SamSaffron:
is that code still even in launcher, I have want to strip it out anyway, ./launcher enter does the trick and people can build a custom template if they really want ssh.

rata:
Oh, ok. No problem if you delete then. But yes,I'd like ssh for now, at least to easily upgrade as unattended upgrades doesn't seemto be working (https://github.com/discourse/discourse_docker/issues/149)

coding-horror:
SSH is not really in here any more.

PR #156, opened 2015-05-06
Fixing a syntax error with one of the exception handlers in mailtest

rosshays:
I ran into an error when using mailtest with one of the exception handlers throwing an error of its own. It was a pretty small change to fix it.


Supermathie:
Whoops. LGTM, thanks!

PR #157, opened 2015-05-08
Added config name check to prevent docker name errors after build.

ericschoolzilla:
It's better to fail early on bad docker names instead of waiting for a build to complete before knowing you've entered a bad docker name.

### DIFF
--- a/launcher
+++ b/launcher
@@ -4,6 +4,16 @@ command=$1
 config=$2
 opt=$3
 
+# Docker doesn't like uppercase characters, spaces or special characters, catch it now before we build everything out and then find out
+re='[A-Z/ !@#$%^&*()+~`=]'
+if [[ $config =~ $re ]];
+  then
+    echo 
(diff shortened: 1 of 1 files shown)


launcher
```diff
@@ -4,6 +4,16 @@ command=$1
 config=$2
 opt=$3
 
+# Docker doesn't like uppercase characters, spaces or special characters, catch it now before we build everything out and then find out
+re='[A-Z/ !@#$%^&*()+~`=]'
+if [[ $config =~ $re ]];
+  then
+    echo 
+    echo "ERROR: Config name must not contain upper case characters, spaces or special characters. Correct config name and rerun $0."
```
rata: Maybe output to stderr ?

(note I'm not the maintainer nor nothing special, just a new user :))

SamSaffron: I think its ok for now, but we should sweep through launcher and output all errors to stderr

SamSaffron:
thanks

PR #161, opened 2015-05-21
Update cipher list

jamielinux:
Direct copypasta from Mozilla's intermediate list. (A few extra ciphers are enabled/disabled.)

RC4 is disabled on Mozilla's list. Enabling RC4 is a mitigation against BEAST for clients using TLS 1.0, but the strength of RC4 has been questioned and some think the trade off isn't worth it. Whether true or not, I think it's better just to stick with Mozilla's widely used list so the PR disables RC4 ciphers.


SamSaffron:
@Supermathie  is this good with you?

Supermathie:
Latest update from Mozilla, eh? Looks good to me.+1

SamSaffron:
cool, merged then.

SamSaffron:
@jamielinux  can you review https://github.com/discourse/discourse_docker/commit/421b9abef0fec6d495d009a82aa1311c43203cf3 to ensure it all looks good

jamielinux:
@SamSaffron Regarding 421b9abef0fec6d495d009a82aa1311c43203cf3, I wonder what the reason is to ignore errors with "|| true". Nginx will fail to start if dhparams.pem doesn't exist. I'll open a PR.

PR #163, opened 2015-05-22
Do not ignore errors when generating DH parameters

jamielinux:
Nginx will fail to start if dhparams.pem doesn't exist.

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -3,7 +3,7 @@ run:
      cmd:
        # Generate strong Diffie-Hellman parameters
        - "mkdir -p /shared/ssl/"
-       - "[ ! -e /shared/ssl/dhparams.pem ] && openssl dhparam -out /shared/ssl/dhparams.pem 2048"
+       - "[ -e /shared/ssl/dhparams.pem ] || openssl dhparam -out /shared/ssl/dhparams.pem 2048"
   - replace:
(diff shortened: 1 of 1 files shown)


SamSaffron:
I think this is cause @riking is ensuring that this "line" will always have a status code of zero even if the file is already from previous bootstrap

riking:
Yeah, if the line has an exit code of 1 the bootstrap will bail (I tested that).

Is there a way to flag the line as "continue if failed" in pups?

SamSaffron:
yes there is actually

https://github.com/discourse/discourse_docker/blob/master/templates/postgres.template.yml#L212

SamSaffron:
though it depends on what fails ... if the check fails cause the file is already there its fine... if generating the file fails, its bad

jamielinux:
Ah I see. `[ ! -e /foo ]` will exit 1 if the file exists, but we also don't want to hide errors in generation. It's probably a vanishingly rare scenario, but openssl might fail to create valid/safe primes and we end up with a dhparams.pem that isn't valid but nginx will happily use.

Does an expanded syntax work for these templates, like `if [ ! -e /tmp/foo ]; then foo; fi`.

riking:
:+1: that makes sense, and I would have used it if I thought of it first

SamSaffron:
thanks!

PR #176, opened 2015-07-28
Removing the mailtest command from the launcher reference

Kovah:
The mailtest command was removed in [579ce85](https://github.com/discourse/discourse_docker/commit/579ce85d19f1b95151126957988122feb6cbbd8b)., should be removed from the `launcher` reference in the README.md file


ZogStriP:
Thanks

PR #177, opened 2015-07-29
Update install docker instructions

krichprollsch:
According with new apt repos description [[1]](https://blog.docker.com/2015/07/new-apt-and-yum-repos/)

### DIFF
--- a/launcher
+++ b/launcher
@@ -195,10 +195,10 @@ install_docker() {
   echo "sudo reboot"
   echo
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
looks good, thanks ! 

We should add a pre-req check that test if lxc-docker is installed and recommend installing docker-engine instead. Would you feel like building something like that ?

krichprollsch:
Yes, I suppose I can do that ^^

SamSaffron:
awesome :)

PR #178, opened 2015-08-14
Add docker_args support to YAML configs (fix #175)

### LINKED ISSUE #175: Can't set --docker-args through YAML file
Because of #174, I've ended up having to run:
```
./launcher start info1-discourse --docker-args '--expose 80'
```
but I can't save those extra options in the `info1-discourse.yml` script, which is inconvenient.

I'd like that to be fixed; that is, I'd like general support for arbitrary options. (Not doing that would only work if discourse_docker made sure to wrap every Docker feature, or addressed every use case of every Discourse user — I believe both are unfeasible.)

Blaisorblade:
Following up on @SamSaffron's https://github.com/discourse/discourse_docker/issues/175#issuecomment-126169578. (I also plan to add some  syntax for `--expose` soon).

### DIFF
--- a/launcher
+++ b/launcher
@@ -401,6 +401,9 @@ run_start(){
    ports=`cat $config_file | $docker_path run $user_args --rm -i -a stdout -a stdin $image ruby -e \
           "require 'yaml'; puts YAML.load(STDIN.readlines.join)['expose'].map{|p| '-p ' << p.to_s << ' '}.join"`
 
+   docker_args=`cat $config_file | $docker_path run $user_args --rm -i -a stdout -a stdin $image ruby -e \
+          "require 'yaml'; puts YAML.load(STDIN.readlines.join)['docker_args']"`
+
    set_template_info
    set_volumes
    set_links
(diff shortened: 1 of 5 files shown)


launcher
```diff
@@ -401,6 +401,9 @@ run_start(){
    ports=`cat $config_file | $docker_path run $user_args --rm -i -a stdout -a stdin $image ruby -e \
           "require 'yaml'; puts YAML.load(STDIN.readlines.join)['expose'].map{|p| '-p ' << p.to_s << ' '}.join"`
 
+   docker_args=`cat $config_file | $docker_path run $user_args --rm -i -a stdout -a stdin $image ruby -e \
+          "require 'yaml'; puts YAML.load(STDIN.readlines.join)['docker_args']"`
```
SamSaffron: isn't this adding a newline here cause its using puts ? does / can this post a problem?

Blaisorblade: TL;DR. Yes, but it works just like in `ports`, so no problem.

In fact, during command substitution (processing backticks) bash chomps [any trailing newlines](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html#Command-Substitution), so the newline doesn't even appear in `$docker_args`.

Removal of trailing newlines explains the following output:
```bash
$ echo foo
foo
$ echo $(echo foo)
foo
$ ruby -e 'puts "foo"'
foo
$ echo $(ruby -e 'puts "foo"')
foo
$ ruby -e 'puts "foo"; puts ""'
foo

$ echo $(ruby -e 'puts "foo"; puts ""')
foo
```

SamSaffron:
nice! thanks

PR #179, opened 2015-08-17
Clarify Ubuntu-specific docker install instructions

Stealthii:
The instructions to install Docker are Ubuntu Precise specific, and are also outdated (recent docker versions require a newer kernel than the now unsupported Raring one). Rather than keep our instructions constantly updated, we should direct the user to the Docker site for installation instructions. I've also updated the example instructions to reflect a recent Ubuntu install, which is our most likely installation target.

### DIFF
--- a/launcher
+++ b/launcher
@@ -187,16 +187,14 @@ get_ssh_pub_key() {
 
 install_docker() {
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
cool, thanks!

PR #187, opened 2015-09-14
Test that we have enough memory and diskspace before installing

mpalmer:
The minimum specs for a Discourse server are a standalone machine with 1GB
of memory + 1GB swap, or 2GB of memory.  To try and catch anyone who's doing
this on a larger machine with lots of other stuff going on, I've checked
"available" memory (used memory minus disk cache) rather than total memory.


SamSaffron:
nice

PR #188, opened 2015-09-14
Add a basic Vagrantfile for local testing

mpalmer:
It comes in handy when you want to poke around with things; anyone who
tries to use this for production-like deployments should be gently but
firmly dissuaded.


SamSaffron:
hmmm, can you add some info to the readme on how to use this, also how is latest version of Docker being installed here?

mpalmer:
Docker isn't being installed automagically.  I'll add that in, and document Vagrant's availability.

mpalmer:
Documentation added, and Vagrantfile improved.

SamSaffron:
cool ... nice one!

PR #191, opened 2015-09-21
Enclose sample smtp password in quotes

riking:
This tells the YML parser to ignore special characters (other than `\` and `"`) inside the quotes.


eriko:
If this does not already apply to database passwords it should.

SamSaffron:
This is not the issue, the pain point is the conf parser in Discourse will
fix.

PR #198, opened 2015-10-09
update elapsed_time

feythin:
`date: invalid date ‘"2015-09-24 02:43:1’`
the container exited date is string, so you should remove the quotes


SamSaffron:
updated to latest per: 


https://github.com/discourse/discourse_docker/commit/4e0c969be01d2b11bcabaefeacfbd35358490570

PR #199, opened 2015-10-15
Trust the real_ip header sent form socket connections

Mooash:
If sockets are being used, trust the `real_ip` header sent form connections to it by default, otherwise all the connections come from the IP `unix` :smile:

### DIFF
--- a/templates/web.socketed.template.yml
+++ b/templates/web.socketed.template.yml
@@ -16,8 +16,10 @@ run:
      from: /listen 80;/
      to: |
        listen unix:/shared/nginx.http.sock;
+       set_real_ip_from unix:;
   - replace:
      filename: "/etc/nginx/conf.d/discourse.conf"
      from: /listen 443 ssl spdy;/
      to: |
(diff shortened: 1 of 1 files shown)


SamSaffron:
sure lets try it, any security implications?

Mooash:
I suppose there are given now if anyone gets access to that socket they can "pretend" to be any IP they want. Although, IMO if someone as access to a socket like that you could have bigger issues on your hands.

All this change does is just trust the `x-real-ip` header sent to the socket.

SamSaffron:
more concerned that an end client can set x-real-ip...

Mooash:
I don't think so, checkout the docs [here](http://nginx.org/en/docs/http/ngx_http_realip_module.html#set_real_ip_from "Module ngx_http_realip_module")

SamSaffron:
I follow that ... but a client can set whatever headers it wants, which means you need to confirm that whatever is upstreaming the headers strips untrusted ones ... so this working securely highly depends on who is sitting between the client and the server.

PR #202, opened 2015-10-16
Add switch to override the default Docker hostname (fixes #193)

### LINKED ISSUE #193: Allow overriding Hostname (Docker -h/--hostname option)
In https://github.com/discourse/discourse_docker/blob/2084bbe7bfb20c3e26c24aaa0ee59e73689fe8a1/launcher#L457, the hostname option for the Docker image is set to `$hostname-$config`. Is there a particular reason for this? 

I'm asking because in my setup, I'd like this to be set to `$hostname`. If you agree, I'll try to post a short Pull Request adding a switch to the config files.

fbender:
A new (optional) ENV var was added to the templates that instructs the `launcher` script to use the DISCOURSE_HOSTNAME as the Docker hostname (via the -h/--hostname option) instead of the default value of `$hostname-$config`.

Successfully tested with `DOCKER_USE_HOSTNAME: true` and with the env variable commented out.

See #193.

### DIFF
--- a/launcher
+++ b/launcher
@@ -502,10 +502,28 @@ run_start(){
    set_run_image
    set_boot_command
 
+   # get hostname and settings from container configuration
+   for envar in "${env[@]}"
+   do
+     if [[ $envar == DOCKER_USE_HOSTNAME* ]] || [[ $envar == DISCOURSE_HOSTNAME* ]]
+     then
+       # use as environment variable
+       eval $envar
+     fi
+   done
+
    (
      hostname=`hostname`
+     # overwrite hostname
+     if [ "$DOCKER_USE_HOSTNAME" = "true" ]
+     then
+       hostname=$DISCOURSE_HOSTNAME
+     else
+       hostname=$hostname-$config
+     fi
+
      set -x
-     $docker_path run $user_args $links $attach_on_run $restart_policy "${env[@]}" -h "$hostname-$config" \
+     $docker_path run $user_args $links $attach_on_run $restart_policy "${env[@]}" -h "$hostname" \
         -e DOCKER_HOST_IP=$docker_ip --name $config -t $ports $volumes $docker_args $run_image $boot_command
 
    )
--- a/samples/standalone.yml
+++ b/samples/standalone.yml
@@ -64,6 +64,10 @@ env:
 
   ## TODO: The domain name this Discourse instance will respond to
   DISCOURSE_HOSTNAME: 'discourse.example.com'
+  ## TODO: Uncomment if you want the container to be started with the same
+  ##       hostname (-h option) as specified above (default "$hostname-$config")
+  ## NOTE: 'true' is the only valid value here, any other will be ignored
+  #DOCKER_USE_HOSTNAME: true
 
   ## TODO: The mailserver this Discourse instance will use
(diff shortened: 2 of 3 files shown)


SamSaffron:
this looks safe @mpalmer  can you give it a quick review as well.

mpalmer:
The proposed approach won't break anything that I can see.  Does it make sense to go a bit further, perhaps?  Create a `CONTAINER_HOSTNAME` parameter that you could set, to make the hostname anything you wanted.  Then, if you want it to be the same as `DISCOURSE_HOSTNAME`, you could set it like that, and if you wanted it to be something else entirely, you could do that too.

SamSaffron:
agree, yeah prefer an optional "CONTAINER_HOSTNAME" env var...

fbender:
Yeah I thought about that, too, however does it ever make sense to use a hostname other than `DISCOURSE_HOSTNAME`? Plus, I wanted to avoid the pitfall that you want to use the same hostname but have a minor spelling error that makes it not quite the same. 

I could, however, allow other values than `true` and `false` or `(missing)`, and use that one as `CONTAINER_HOSTNAME` (and as before, if the value is `true`, use `DOCKER_HOSTNAME` as the container hostname). I was planning on doing that in a later step (since I wanted to get the minimal viable solution to work first, and the answer to the question above is probably "yes"), so if you agree, please accept this PR and I will prepare a new one for the extended functionality.

SamSaffron:
I think this is treading into the world of what are better defaults:

```
root@www%
```

vs

```
root@tiefighter2-wd
```

The latter provides a hint on machine the container is hosted on which is
very handy for multi host deploys.

But... on digital ocean out of the box `./launcher enter app` is displaying

```
root@discourse
```

if the hostname of the parent is: `discourse.site.com` ... so I think the
fix we need FIRST

is to change it so the hostname becomes:

discourse-app.site.com ... instead of discourse.site.com-app, this will at
least ensure you have a sane prompt.

Can you do a PR for that first.

fbender:
Like so: fbender/discourse_docker@e0d9ceffe4e32c80f5df7fc26db4e256cba99eeb ?

I implemented this on top of my changes in this PR.

fbender:
@SamSaffron did you have a chance to look at the commit, yet?

SamSaffron:
sorry about the delay, I have some family issues at the moment so my
schedule is all over the place, will have think about this and sort it out.

fbender:
Sure, no worries, take your time!

fbender:
In the meantime, do you have any comments @mpalmer ?

coding-horror:
this change still seems good to me, can you rebase @fbender ?

fbender:
Sure, will do so in the coming days.

coding-horror:
still needs rebase @fbender ;)

fbender:
Uh, yes, forgot about it. Sorry! Scheduled for tomorrow morning.

fbender:
Rebased! 

The additional change requested by @SamSaffron is available in the [`change-default-docker-hostname`](https://github.com/fbender/discourse_docker/tree/change-default-docker-hostname) branch. Is that commit how you envisioned it? Shall I merge it into this branch?

SamSaffron:
yes I think this is fine, thanks!

PR #205, opened 2015-11-18
Install pngcrush 1.7.87 instead of 1.7.85

dereckson:
pngcrush 1.7.85 isn't downloadable anymore

According to http://sourceforge.net/projects/pmt/files/pngcrush/
only two versions are directly available: the released 1.7.87 and up-to-come 1.7.88.

**Upstream changelog**

pngcrush-1.7.87 fixes a double-free bug in the sPLT chunk handling


SamSaffron:
thanks

PR #208, opened 2015-12-01
Replace SPDY directives with HTTP/2

Stealthii:
The upstream image we build from is using a newer, mainline build of Nginx, which no longer supports SPDY. Support for HTTP/2 has been added however, and this PR updates the configuration to reflect this change.

Tested and working on https://discourse.farsetlabs.org.uk/


SamSaffron:
wow nice!

PR #211, opened 2015-12-07
FEATURE: Add check to ensure Docker server and client versions are in…

tgxworld:
… sync.

Meta: https://meta.discourse.org/t/docker-gc-version-mismatch/36382

### DIFF
--- a/launcher
+++ b/launcher
@@ -8,7 +8,7 @@ opt=$3
 re='[A-Z/ !@#$%^&*()+~`=]'
 if [[ $config =~ $re ]];
   then
-    echo 
+    echo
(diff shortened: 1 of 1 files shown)


tgxworld:
@SamSaffron  updated :smile:

coding-horror:
did we want to merge this?

SamSaffron:
I don't know how I feel about this, lets leave it for now. 

The actual issue that was raised was cause we were allowing docker 1.5 servers which we no longer support. 

If it pops up again we can look at merging, just worry it will stop people from installing in some situations where it really does not matter.

PR #213, opened 2015-12-21
Update TLS ciphers

jamielinux:
A few extra ciphers were added to the Mozilla wiki list.


SamSaffron:
@mpalmer are we good with this (if yes can you merge it in)

mpalmer:
I'm not good with this change -- there's no rationale I can find for why these additional suites are required.  Which clients need these specific suites, and not any of the ones already in the list (or any other, stronger, cipher)?  (That's not directed at you, specifically, @jamielinux; I've dug into the commits on the Mozilla wiki, and I can't find any useful info).

Evaluating the default ciphersuite choice further, the set of browsers which need this "Intermediate" compatibility list are not browsers which Discourse even *tries* to support.  Should we switch to the "modern" ciphersuite list (or, preferably, the suite list at https://cipherli.st/ which is much shorter and easier to audit)?

SamSaffron:
Your call Matt, totally fine with whatever you think is right.

jamielinux:
@mpalmer Hmm. On a closer inspection, the two additional 3DES ciphers (`ECDHE-RSA-DES-CBC3-SHA` and `ECDHE-ECDSA-DES-CBC3-SHA`) are for browsers that Discourse doesn't support, so feel free to close this PR. [i](https://wiki.mozilla.org/Security/Server_Side_TLS#RC4_weaknesses) [ii](https://bugzilla.mozilla.org/show_bug.cgi?id=927045)

> Should we switch to the "modern" ciphersuite list

Discourse browser support (Safari 5.1+, Chrome 23+, IE 10+, Firefox 16+, Android 4.1+) lies between the "modern" and "intermediate" compatibility lists. The "modern" list would require bumping Discourse's requirements (Safari 7+, Chrome 22+, IE11+, Firefox 27+, Android 4.4+). Chrome isn't affected at all, and usage of the dropped Firefox/Safari versions is pretty minimal. But it drops support for [27% of Android users](http://developer.android.com/about/dashboards/index.html) (4.1 to 4.3) and something like 9% of IE users (IE10).

> or, preferably, the suite list at https://cipherli.st/ which is much shorter and easier to audit

The Mozilla TLS page is well maintained by a more diverse set of contributors, including Mozilla employees. It has a clear browser support matrix and is generally more thorough.

My view is that the "modern" list supports too few browsers. The "intermediate" list may support more browsers than Discourse supports, but I imagine trying to slice the list down to match the current Discourse browser support matrix is more effort than it's worth.

PR #215, opened 2016-01-07
Using docker link over port in multi container samples

fantasticfears:


samples/web_only.yml
```diff
@@ -28,8 +33,8 @@ env:
   DISCOURSE_DB_SOCKET: ''
   #DISCOURSE_DB_USERNAME: discourse
   DISCOURSE_DB_PASSWORD: SOME_SECRET
-  DISCOURSE_DB_HOST: DB_IP_ADDRESS
-  DISCOURSE_REDIS_HOST: REDIS_IP_ADDRESS
+  DISCOURSE_DB_HOST: DB_CONTAINER_ALIAS
+  DISCOURSE_REDIS_HOST: REDIS_CONTAINER_ALIAS
```
SamSaffron: hmm how is this going to work? who is setting DB_CONTAINER_ALIAS ? 

if only a container named "data" is being linked it would probably simply be "data" there.

fantasticfears: Good point. Updated

SamSaffron:
much better! thanks

wbond:
Was it really intended to remove sshd functionality from the "standard" template (https://github.com/discourse/discourse/blob/master/docs/INSTALL-cloud.md) like this? I spent hours today trying to figure out why my automated deploy of Discourse was all of a sudden breaking in an odd way.

Specifically, `./launcher ssh app` no longer works, and anything that used SSH to the docker container is now also broken.

Which then leads me to ask if there is any intention of having real "releases" of discourse_docker in the future? Since this is the only supported way to install Discourse, it would be nice if there were some versioned tags so scripted setups could target a known, working version of docker_discourse.

fantasticfears:
The usage of `sshd` is removed from sample files. The old configuration asked to use `sshd.template.yml` still works. I don't know why yours is not working.

If your automated bootstrap system copy the files from the samples, you do have a good reason to change the code. As for discourse_docker, there's always a good reason leading the code to evolve. Old samples [exposed](https://meta.discourse.org/t/losing-connection-with-redis/36343) the port to the host, this commit fixs this pontential security problems.

Although, if you want to control all of these procedures and make sure they work even after the upstream is updated, creating your own samples and templates is not a bad idea.

The new configuration and recommend way to enter the container is `./launcher enter app`

wbond:
Clearly `sshd` was removed from the sample files. My process stopped working because my configuration scripts used the sample and then modified variables I needs changes, just like the official install docs outline.

My point is that this change "breaks" documented features outlined in the one, official, supported install method, without even mentioning it anywhere (commit message, docs, etc). The documentation in this repo still says `./launcher ssh app` is a feature, whereas after this change it is not unless you figure out what templates are, that there is a special one for ssh and that you have to manually add it in.

I know well that features must change over time. My question is more of a: will you consider explicit "releases" of discourse_docker? Most projects have a versioning strategy where certain version number changes signify backwards compatibility breaks. Also changelogs. This was a pretty significant change without a commit message or documentation change that anywhere mentions removing the (previous) default ssh support.

If I hardcode a sample file but then features in the repo change, I'm trading one way for things to break for another.

Also, `./launcher enter app` may be great for some things, but isn't really useful when using a tool like Ansible for automated configuration.

For now, I've pinned the repo to a previous commit to get a "stable" version.

Hopefully my comments here won't come off as something you need to rebut, but rather feedback about how people who are consuming this project could benefit from some workflow/process tweaks.

SamSaffron:
I get it, but we got to fix the documentation (I just tried), ssh in the
container is a ninja option now, even at scale we do not use it and prefer
to use `docker exec` it is just way too fragile and not worth carrying when
docker exec gets us all the way there 99% of the time

PR #216, opened 2016-01-10
Kill runsvdir by sending a TERM.

tgxworld:
meta: https://meta.discourse.org/t/runsv-hanging-on-docker-container-shutdown/36844/16


tgxworld:
@SamSaffron We should be able to send it a TERM signal safely since we've already shut down all the services.

SamSaffron:
I think I will go with tini for the next image, waiting on some feedback
first.

tgxworld:
Hmm if there is not shut down script for some of our image we'll not be able to send it a `TERM`

PR #219, opened 2016-02-14
default the lets encrypt directory to the standalone setup

renandincer:
mentioned this briefly [here](https://meta.discourse.org/t/support-for-lets-encrypt/22308/51?u=renan) on the meta discourse.

This changes the default setup assumed in the lets encrypt template. The previous configuration assumed that a user setting up letsencrypt uses a multi-container setup. However, we should push new users (who start with the standalone setup) to use SSL/HTTPS more often, easier.

Also I'd like to bring up either 
1. An update to the SSL tutorial to make it easier to set up encryption freely. 
or maybe in the future:
2. SSL on by default on a new standalone docker installation using lets encrypt.

what do we think? too much information shared with lets encrypt?

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -1,5 +1,5 @@
 env:
-  LETSENCRYPT_DIR: "/shared/letsencrypt"
+  LETSENCRYPT_DIR: "/shared/standalone/letsencrypt"
 
 run:
   - exec:


tgxworld:
Hi @renandincer! Thanks for the PR :). However, I don't quite understand why this change is needed. The current configuration assumes a single container setup actually. 

In `standalone.yml`: 

```
volumes:
  - volume:
      host: /var/discourse/shared/standalone
      guest: /shared
```

The `/shared` directory in the container is mounted onto the host at `/var/discourse/shared/standalone`.  That means `/shared/letsencrypt` will be at `/var/discourse/shared/standalone/letsencrypt` on the host.

PR #220, opened 2016-02-18
Added link to Meta howto to standalone sample.

fefrei:
see https://meta.discourse.org/t/running-other-websites-on-the-same-machine-as-discourse/17247/74?u=fefrei


SamSaffron:
thanks

PR #225, opened 2016-03-17
Add a simple setup ruby script

sethherr:


sethherr:
... not ready yet

SamSaffron:
I am very very strongly against requiring ruby on the host. 

Open to possibly invoking more of launcher in ruby via disposable containers. But requiring ruby on host is out of the question.

PR #226, opened 2016-03-19
FEATURE: change login directory to discourse home when enter a container

fantasticfears:
Although rails commands automatically exposed in the environment in a container, it's easier to manipulate files, e.g. I update my plugins (no client assets) to minimize rebuild time.
I think it's better to point the user to the `/.

If chosen `.bach_profile`, an interactive bash is needed.

### DIFF
--- a/launcher
+++ b/launcher
@@ -576,7 +576,7 @@ case "$command" in
       ;;
 
   enter)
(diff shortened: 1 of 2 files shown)


coding-horror:
is this OK @SamSaffron

SamSaffron:
sure this is fine

PR #227, opened 2016-03-23
Add a simple_setup command to launcher

sethherr:
- Get user configuration input before running `prereqs`, `simple_setup` and bootstraping so the user can enter input immediately and everything runs after
- Make a simple_setup ruby script that updates the configuration file without destroying the comments
- Use throwaway discourse docker images to run simple_setup
- Add a ruby_scripts folder and make it a data volume in run_ruby_script, to make running ruby easier, refactorable


launcher
```diff
@@ -413,6 +414,13 @@ else
   restart_policy=${restart_policy:---restart=always}
 fi
 
+run_ruby_script() {
+  ruby_file=$1
+  arguments=$2
+  cp containers/"$config".yml ruby_scripts/config.yml  # Add config to the data volume for easy access
+  echo `$docker_path run $user_args --rm -i -a stdout -a stdin -v /var/discourse/ruby_scripts:/ruby_scripts $image ruby /ruby_scripts/"$ruby_file" "$arguments"`
```
SamSaffron: this is forcing launcher to be installed in /var/discourse ... can not have this tight dependency, use relative to pwd paths

launcher
```diff
@@ -52,16 +52,17 @@ fi
...
+  echo "    stop:         Stop a running container"
+  echo "    restart:      Restart a container"
+  echo "    destroy:      Stop and remove a container"
+  echo "    enter:        Use nsenter to enter a container"
+  echo "    ssh:          Start a bash shell in a running container"
+  echo "    logs:         Docker logs for container"
+  echo "    bootstrap:    Bootstrap a container for the config based on a template"
+  echo "    rebuild:      Rebuild a container (destroy old, bootstrap, start new)"
+  echo "    cleanup:      Remove all containers that have stopped for > 24 hours"
+  echo "    simple_setup: Configure, bootstrap and start Discourse"
```
SamSaffron: I would call this "setup", no need to call it "simple_setup"

launcher
```diff
@@ -513,8 +521,50 @@ run_start(){
...
+    DISCOURSE_SMTP_ADDRESS
+    DISCOURSE_SMTP_PORT
+    DISCOURSE_SMTP_USER_NAME
+    DISCOURSE_SMTP_PASSWORD
+  )
+  config_variables_desc=(
+    'Domain name'
+    "Initial admin and developer email(s)"
+    '(email provider) SMTP Address'
+    '(email provider) SMTP Port'
```
SamSaffron: certain stuff should default, in particular ... port

SamSaffron:
Overall this change is good, can you do a screen grab of a demo run?

sethherr:
And a screenshot:

screen shot 2016-03-24 at 10 49 45 am

coding-horror:
any other comments on this @SamSaffron ?

SamSaffron:
Why do we need to carry `ruby_scripts/config.yml` file around?

sethherr:
If it's in the docker data volume it's easy to read and modify it. This was the simplest idea I came up with.

There are 11 places in the launcher right now that read the config file into stdin to parse it with (inlined) ruby.

I figured I'd submit a pr refactoring those into the `run_ruby_script` function if this was accepted and maybe write some tests.

SamSaffron:
I just feel that now config ends up living in 2 spots ...once in the "containers" folder and another in the "temp" folder. It feels somewhat odd. 

Can't it simply use the existing config file it generated to figure out the current settings? (eg: app.yaml) anything tmp can go in `/tmp`

sethherr:
What if I moved the ruby scripts into `containers` and made `containers` the data volume? 

Then the config file wouldn't be duplicated and it would retain all the simplicity I was gunning for.

coding-horror:
we are still looking at this -- Sam just said we still do want to move the launcher logic into Ruby.

sethherr:
Just tell me what direction you'd like to go, if there is anything I can do to help

coding-horror:
@sethherr can you take a look at the way it is now? We restructured into two Bash scripts, one that is "discourse specific" (cough @SamSaffron) and one that is "just launch a docker container"..

the idea is that it's easier for newbies if they don't have to run an editor at all but can just answer some prompts at the command line and have most of the magic happen.

This does nothing for getting us out of hideous-ancient-Bash-scripting-language territory, but it _does_ remove several points of failure for people trying to install Discourse (YAML file errors, "how do I use Nano", and so on).

PR #228, opened 2016-03-24
Template for serving through an .onion address.

RomeoPapaSK:
Thread on meta.discourse.org:
https://meta.discourse.org/t/template-for-serving-through-an-onion-address-with-docker

It creates a new `onion.conf` file in the nginx config folder. The new `onion.conf` is created by copying `nginx.sample.conf` and editing that copy.

This new config file creates another site on nginx's port 80, that listens to requests made to the specified .onion address.

The template checks that a `DISCOURSE_ONION` ENV variable is defined.

### DIFF
--- a/templates/web.onion.template.yml
+++ b/templates/web.onion.template.yml
@@ -1,0 +1,49 @@
+# Adds another server on port 80 for hidden service hosting
+
+run:
+  - exec:
+      cmd:
+        # Check DISCOURSE_ONION variable has been configured
+        - if [ -z "$DISCOURSE_ONION" ]; then echo "DISCOURSE_ONION ENV variable is required and has not been set."; exit 1; fi
+
+  - exec:
+      cmd:
+        # Copy default nginx file
+        - "cp $home/config/nginx.sample.conf /etc/nginx/conf.d/onion.conf"
+
+  # Remove duplicate entries that would crash the server
+  - replace:
+      filename: "/etc/nginx/conf.d/onion.conf"
+      from: /upstream[^\}]+\}/m
+      to: ""
+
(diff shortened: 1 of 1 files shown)


RomeoPapaSK:
Thread on meta.discourse.org:
https://meta.discourse.org/t/template-for-serving-through-an-onion-address-with-docker

SamSaffron:
Can you add a comment to the top to link to the meta howto?

Also, why add an additional conf file, why not simply change the existing one to also support onion?

RomeoPapaSK:
Link moved to top.

> Also, why add an additional conf file, why not simply change the existing one to also support onion? 

The reason it adds an additional conf file is basically because of `web.ssl.template.yml`. Depending on whether it's loaded or not, it will patch the content of `discourse.conf`, and that would mean I'd need to make two different patches: one for when the ssl.template is applied and one for when it's not applied.

I could also have only one patch, working on discourse.conf but we have to make sure they include the onion.template *before* the ssl.template in `app.yml`. That might imply modifications to `web.ssl.template.yml` so the patch doesn't get confused between the two `server` definitions.

This is why I chose to go whith an additional conf file. It's just simpler: copies from the .sample.conf and doesn't have to worry about which templates have been applied or not.

I could however get the name of the conf file to change from `onion.conf` to `discourse-onion.conf`.

SamSaffron:
this sounds reasonable, still we need a comment at the top of the file explaining usage and linking to meta... 

merging this in for now! thank

RomeoPapaSK:
Cool.

OK I'll send you a pull request with a comment at the top of file explaining usage and linking to meta.

PR #230, opened 2016-03-29
Lets Encrypt Should Become Default

djtye:
Lets make Discourse have Lets Encrypt by default.


djtye:
Done a few changes :)

PR #233, opened 2016-04-19
Assign a stable but unique MAC address to each container

mpalmer:
This does Very Good Things for IPv6 addresses (at least on versions of
Docker that aren't [buggy](https://github.com/docker/docker/issues/17739)).
You get an IPv6 address which doesn't change on each reboot, and at the same
time you don't get different containers having the same IPv6 address at
different times.  Wins all round!

### DIFF
--- a/launcher
+++ b/launcher
@@ -528,9 +528,12 @@ run_start(){
      # docker added more hostname rules
      hostname=${hostname/_/-}
 
+     mac_address="--mac-address $(echo $hostname | md5sum | sed 's/^\(..\)\(..\)\(..\)\(..\)\(..\).*$/02:\1:\2:\3:\4:\5/')"
+
      set -x
      $docker_path run $user_args $links $attach_on_run $restart_policy "${env[@]}" -h "$hostname" \
-        -e DOCKER_HOST_IP=$docker_ip --name $config -t $ports $volumes $docker_args $run_image $boot_command
+        -e DOCKER_HOST_IP=$docker_ip --name $config -t $ports $volumes $mac_address $docker_args \
+        $run_image $boot_command
 
    )
    exit 0
--- a/launcher
+++ b/launcher
@@ -528,7 +528,7 @@ run_start(){
      # docker added more hostname rules
      hostname=${hostname/_/-}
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
I may hold off until this is triaged: 

https://github.com/docker/docker/issues/22174

mpalmer:
Don't make me do Birthday Paradox math... you wouldn't like me when I have to do math...

Dammit, now I've done math.

MD5 has an excellent waterfall (as you'd expect from a cryptographic hash function).  So we can safely assume that all hashes are independent events.  Thus, we can use the usual methods (as per [this page](http://preshing.com/20110504/hash-collision-probabilities/) to calculate out the probability of, say, any one of 100 different container names (and I doubt anyone's running 100 `launcher`-managed containers on a single machine) having the same 40-bit hash value.  It comes out to be about 1 in 222 million, which according to the handy chart in the above-linked page, comes out to be just a bit more likely than dying in a shark attack.  It's probably significantly *less* likely than an Australian dying in a shark attack, and I don't about you, but I don't worry too much about dying in a shark attack, so this probably shouldn't worry either of us, either.

Do I dare awaken the beast by comparing the probability of a MAC address collision with the probability of a recoverable error in a non-ECC RAM-using system?  No.  No, I do not.

coding-horror:
Both are extremely unlikely 🙃

coding-horror:
so, should we merge this, or no?

SamSaffron:
Ill merge it and watch it ... 

@mpalmer  for safety can you run that code in a container so we know we are getting the right version of sed / md5sum (not sure it is called md5sum everywhere or installed everywhere) etc... been bitten on stuff like this in the past

mpalmer:
I *can* run it in a container, but I don't think it's necessary.  The features of `sed` I'm using are all in POSIX (so anything which doesn't support them is unhinged), and `md5sum` is part of `coreutils` (and also implemented in busybox), which also provides such esoteric stuff as `cat` and `cp`.  Anyone who's mucking around with those probably isn't going to get very far running the rest of launcher, either.

mpalmer:
Extra commit added.

PR #236, opened 2016-04-20
Add memconfig to launcher: Configure defaults based on available RAM

pfaffman:
Creates or modifies config of db_shared_buffers, db_work_mem, and
UNICORN_WORKERS based on available RAM. Parameters not set in config
file (i.e., commented out), are left unchanged and a warning is printed.
Uses samples/standalone_template.yml as basis, but if accepted, I
recommend uncommenting these defaults in standalone.yml and using it
instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Commands:
     destroy:    Stop and remove a container
     enter:      Use docker exec to enter a container
     logs:       Docker logs for container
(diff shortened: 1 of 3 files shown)


coding-horror:
Ok let's try this, since it is behind a command line switch for now it is safe.

PR #237, opened 2016-04-20
Add mail config sanity check

pfaffman:
Checks that all SMTP config variables are set. If not, a warning is
printed and user gets option to ^C to fix.


coding-horror:
OK cool let's try this I have an artisanal install scheduled for about 30 minutes from now and I can test this.

PR #247, opened 2016-04-27
Moved parameter test back before shift

codetaylor:
If the less than test is placed after the shift, usage will be shown even if two parameters are passed.

### DIFF
--- a/launcher
+++ b/launcher
@@ -21,15 +21,14 @@ usage () {
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


tgxworld:
This has been fixed in https://github.com/discourse/discourse_docker/commit/65573a0ea56dff279216884dacfad608432242d7. Thank you for your PR though :+1:

PR #248, opened 2016-04-27
Produce random-but-stable MAC addresses, take 2

mpalmer:
This version works properly on ye olde versions of Docker (v1.7 was reported
as not working with the previous version of this patch).


coding-horror:
I guess you can review @SamSaffron since the first one was not quite right?

SamSaffron:
this looks right to me

PR #250, opened 2016-04-28
Launcher copyedit and discourse-setup tweaks

pfaffman:
Spurious "a" at the end of a line in `launcher`. 

## `discourse-setup` Tweaks:

- give people a chance to ^C before running bootstrap
- Tell people who don't meet config requirements to edit config and run bootstrap themselves.
- Use && to skip starting app if  bootstrap bootstrap failed
- Bonus: whitespace removal!

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -372,6 +372,8 @@ validate_config() {
 
   if [ "$valid_config" != "y" ]; then
     echo -e "\nSorry, these $config_file settings aren't valid -- can't continue!"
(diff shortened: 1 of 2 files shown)


coding-horror:
Why is this extra pause needed? They already get a pause in "does this look right?" so they could just CTRL+C there as well.

Also we have a policy of not accepting whitespace / formatting checkins, and it looks like there is a lot of that in this checkin.

coding-horror:
making some of the changes manually, since a lot is bundled here..

PR #256, opened 2016-05-13
COPYEDIT: Default SMTP port is 25, not 587

pfaffman:
`standalone.yml` indicates that default SMTP port is 587, but it's really 25.

### DIFF
--- a/samples/standalone.yml
+++ b/samples/standalone.yml
@@ -15,7 +15,7 @@ templates:
 ## Uncomment these two lines if you wish to add Lets Encrypt (https)
   #- "templates/web.ssl.template.yml"
   #- "templates/web.letsencrypt.ssl.template.yml"
- 
+
 ## which TCP/IP ports should this container expose?
 ## If you want Discourse to share a port with another webserver like Apache or nginx,
 ## see https://meta.discourse.org/t/17247 for details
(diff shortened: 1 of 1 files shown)


eriko:
It is if you are doing Secure SMTP (SSL / TLS) which I think is why it is set that way.

tgxworld:
Chances are people are not running their own mail server so I think the default port of 587 is corrrect

> [Port 587: 
This is the default mail submission port. When a mail client or server is submitting an email to be routed by a proper mail server, it should always use this port.](http://blog.mailgun.com/25-465-587-what-port-should-i-use/)

coding-horror:
when the default of `DISCOURSE_SMTP_ENABLE_START_TLS` = true, then it implies 587, whereas when `DISCOURSE_SMTP_ENABLE_START_TLS` = false, it implies 25.

So these settings are connected. 587 default _is_ correct if you also use the default value of `DISCOURSE_SMTP_ENABLE_START_TLS` which is **TRUE**.

coding-horror:
Well, looking further, I'm not sure. Perhaps @SamSaffron can clarify.

coding-horror:
OK! Resolved in this commit, which also adds smarter mailgun username handling too. 587 is the explicit value now, we won't rely on defaults any more.

https://github.com/discourse/discourse_docker/commit/7c2777f949553efb541d930f460955fff2040369

PR #261, opened 2016-05-26
Fixed paths in the postgres upgrade instructions

fefrei:


SamSaffron:
sure, this looks right

PR #266, opened 2016-06-30
Fix engine substitution in phpBB3 importer template

gschlager:
The engine substitution (always use MyISAM instead of InnoDB) didn't work anymore because MariaDB 10.1.7+ sets `SQL_MODE` to `NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION` by default.


SamSaffron:
k lets try it

PR #275, opened 2016-08-08
Segregate postgres upgrade instructions on version

lbergnehr:
The previous instructions always instructed the reader to undo the base image and postgres template in the container configuration. If the upgrade were done from 9.3, that might be confusing as the `base_image` property had never been altered in a previous step.


SamSaffron:
Sure, we can try this out

PR #290, opened 2016-09-12
FIX: make swap when mem is <= (not <) 2GB

pfaffman:


pfaffman:
Wait. Needs one more fix.

PR #292, opened 2016-09-20
Improve image building (primarily for discourse_dev)

JaredReisinger:
Move "discourse" user and "/var/www" directory creation into base image so that discourse_dev can build from base image directly instead of requiring the overhead of the full discourse image.

Remove discourse_fast_switch dependency from discourse_dev, and remove the ruby 2.2/2.0 logic, as Discourse already appears to require 2.3.0 or
greater.

Remove discourse_dev's independent `postgres.template.yml` and `redis.template.yml` files, and instead derive them from the top-level templates directory.  This ensures that the discourse_dev image is always up-to-date with the discourse main image.

Add password-less "sudo" permissions for the "discourse" user in the discourse_dev image.  This simplifies (fixes) the "bundle install" command (and others) as implemented by discourse/bin/docker utilities.

Add image/Makefile so that images can be built even on machines without Ruby installed.  Logic is replicated from `image/build.rb`, and should result in equivalent images.  (Perhaps build.rb should be removed?)

Add automation in Makefile to ensure that all Dockerfiles are updated to match the version number in the Makefile.  (This includes the `/VERSION` file written into the base image.)

Add `image/README.md` to describe the images and process.

Update ImageMagick to 6.9.5-9 in the base Dockerfile because 6.9.5-8 is no longer available. (Perhaps this dependency should be vendored.)

Fix base "boot" failure caused by missing log files on a new instance (in `base/runit-1.d-00-fix-var-logs`).  Before chowning the log files, "touch" them to ensure they exist.  This failure previously prevented the discourse_dev `ensure-database` script from running, which made development use of discourse_dev very difficult.

Update to version 1.3.7.  (Driven by IMAGE_VERSION in the Makefile. Mostly, this is to ensure that the next official build of the images will be *at least* 1.3.7, superseding any previous versions.)


JaredReisinger:
Ahhh... one other note.  Prior to this PR, there's a naming discrepancy between the image directories and the actual image names:
* _directory -> image (comment)_
* base -> base (identical, but no "discourse_" prefix)
* discourse -> discourse
* discourse_dev -> dev (drops "discourse_" prefix)
* discourse_test -> discourse_test (keeps "discourse_" prefix)

I made the unilateral decision that the directory name should be used when creating the image name, so now all of the images other than "base" start with "discourse", and _**all**_ of them match their directory names.  For consistency, it might make sense to either change "base" to "discourse_base", or remove the "discourse_" prefix from all of the other directories/images.

JaredReisinger:
... and for docker-squash on OSX, I just saw that docker-squash has a dependency on gnu-tar 1.27, while I have OSX-default bsdtar 2.8.3.  I'm trying "brew install gnu-tar" to see if that does the trick, and will reïnstate docker-squash for OSX if it works.  (But IMO that can easily be a later PR/enhancement of everything else here is okay.)

SamSaffron:
In general I do all my builds on Linux, we don't even need to support building of images on osx.

JaredReisinger:
I figured that was likely the case... but it was sure handy for me to be able to test things out on my dev machine!  It looks like gnu-tar plus either some PATH trickery (or a pending docker-squash change to use 'gtar' on OSX) will get this working.  But it definitely falls in the "nice-to-have" bucket (rather than "must-have") in my book.

SamSaffron:
did you test the latest docker squash, old squash was totally broken on 1.10 but there is a fork out there that fixes it

JaredReisinger:
What finally worked for me (_after_ submitting the PR) was docker-squash 0.2.0 (the one acquired through the Makefile, from https://github.com/jwilder/docker-squash/releases/download/v0.2.0/docker-squash-darwin-amd64-v0.2.0.tar.gz), but with 'gtar' replacing 'tar' on my path, and making sure to run as root.  Note that I haven't yet tried _using_ that image, though... but docker-squash didn't exit with an error code.

I haven't delved into docker-squash enough to know _why_ running as root is required, but it gives me the willies.  If there were a way to mount a working directory with the not-yet-squashed .tar in a container and run docker-squash _inside_ the container (and let it write the squashed tar in that same mounted volume), that might actually take care of both issues: the 'tar' inside the container could be the required one, and running as "root" inside a container isn't quite so scary.

Interestingly, it looks like the long and winding discussion to get squashing built in to Docker may be coming to an end: https://github.com/docker/docker/pull/22641  (But I'm not holding my breath, it was opened in May, and still isn't merged, even as an "experimental" feature.)

image/base/Dockerfile
```diff
@@ -121,3 +120,8 @@ ADD boot /sbin/boot
 ADD cron /etc/service/cron/run
 ADD rsyslog /etc/service/rsyslog/run
 ADD cron.d_anacron /etc/cron.d/anacron
+
+# Create discourse user and /var/www location for both
+# discourse and dev images.
+RUN useradd discourse -s /bin/bash -m -U &&\
```
SamSaffron: Ideally I don't want this user created in base, base can be used by non-discourse projects

image/base/install-imagemagick
```diff
@@ -19,7 +19,7 @@ make all && make install
 
 
 # Build and install ImageMagick
-wget -O $WDIR/ImageMagick.tar.gz "http://www.imagemagick.org/download/ImageMagick-6.9.5-8.tar.gz"
+wget -O $WDIR/ImageMagick.tar.gz "http://www.imagemagick.org/download/ImageMagick-6.9.5-9.tar.gz"
```
SamSaffron: this is fine

image/base/runit-1.d-00-fix-var-logs
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 mkdir -p /var/log/nginx
 chown -R www-data:www-data /var/log/nginx
-chown www-data:www-data /var/log/nginx
```
SamSaffron: you sure about this? I think it is fine but need confirmation

image/discourse_bench/Dockerfile
```diff
@@ -17,11 +19,12 @@ RUN gem install facter &&\
 
 # reconfigure PostgreSQL template databases to use UTF-8
 # otherwise rake db:create fails
-RUN pg_dropcluster --stop 9.3 main &&\
-    pg_createcluster --start -e UTF-8 9.3 main
+RUN pg_dropcluster --stop 9.5 main &&\
```
SamSaffron: cool happy about this, did you test the discourse bench image?

image/base/Dockerfile
JaredReisinger: Fair enough... In that case it may make sense to create an image/layer between base and discourse & discourse_dev.

SamSaffron: I think that is overkill, having a couple of lines of duplication here is not really the end of the world imo

image/base/runit-1.d-00-fix-var-logs
JaredReisinger: Yes... It's completely redundant given line 3.  `/var/log/nginx` and it's descendents are chown'd, so all line 4 does is reïterate that `/var/log/nginx` is supposed to be chown'd.

image/base/Dockerfile
JaredReisinger: Glad you reconsidered.  :-)  I was about to come back playing devil's advocate about YAGNI and premature optimization.

image/discourse_bench/Dockerfile
JaredReisinger: I've done both of these:

```
docker run --rm discourse/discourse_test:1.3.7
docker run --rm discourse/discourse_bench:1.3.7
```

The tests all passed, and bench got through all three benchmarks (both as admin and not), but then threw an EACCES exception (could not open `/root/.puppetlabs/opt/facter/facts.d`).  I'll poke around and see if there's a missing step in the discourse_bench Dockerfile or something.

(As a later change—unless you really want it right now—I could add "test" and "bench" targets that do the above commands.  They could even be them dependencies of "all" so that they are run by default.)

JaredReisinger: Got it!  I'll have a PR into discourse to tweak `scripts/bench.rb` so it stops attempting to load unused external facts.  With that change in place the benchmark runs successfully on my machine (via discourse/discourse_bench):
```
Your Results: (note for timings- percentile is first, duration is second in millisecs)
---
categories_admin:
  50: 94
  75: 100
  90: 108
  99: 130
home_admin:
  50: 98
  75: 104
  90: 115
  99: 145
topic_admin:
  50: 49
  75: 50
  90: 53
  99: 71
categories:
  50: 52
  75: 55
  90: 60
  99: 78
home:
  50: 60
  75: 63
  90: 73
  99: 94
topic:
  50: 18
  75: 18
  90: 19
  99: 35
timings:
  load_rails: 5266
ruby-version: 2.3.1-p112
rss_kb: 269892
pss_kb: 261193
operatingsystem: Ubuntu
virtual: docker
architecture: amd64
physicalprocessorcount: 4
kernelversion: 4.4.20
processor0: Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
memorysize: 1.95 GB
```

JaredReisinger: https://github.com/discourse/discourse/pull/4454 allows running the benchmarks via discourse/discourse_bench docker container.

image/discourse_bench/Dockerfile
```diff
@@ -17,11 +19,12 @@ RUN gem install facter &&\
...
+    pg_createcluster --start -e UTF-8 9.5 main
 
 # create role "discourse"
 RUN /etc/init.d/postgresql start &&\
+    sleep 10 &&\
```
SamSaffron: sleep 10 seems a lot, no?

image/base/Dockerfile
SamSaffron: one moment there :) are you moving these like around to other 2 images?

image/discourse_bench/Dockerfile
JaredReisinger: Yeah... you're right.  I remembered seeing a "sleep 10" somewhere, but it was for redis.  Postgres has "sleep 5" in several places, so I switched to that instead.

SamSaffron:
Hi @JaredReisinger I am back from my break and ready to merge this in, only bit of feedback that needs addressing is moving the "discourse" account creation out of "base" into the various images

PR #297, opened 2016-10-11
Fix OS X not knowing the free bash command

simongareste:
Starting to play with Docker and discourse, developing on a mac, the `free` bash command is not available (to figure out memory availability). This commit fixes that.


SamSaffron:
looks good to me! 

Thanks

PR #301, opened 2016-10-18
Dev image improvements

xfalcox:
This builds on awesome work by @JaredReisinger and make some adjustments.

Still some ways to go.


xfalcox:
@SamSaffron I think this is good enough for a review.

This is phase 1, we need work on the others steps still.

SamSaffron:
This looks good to me... lets get an image published! 👍

PR #303, opened 2016-10-28
Fix physical core calculation

theikkila:
Related to this thread https://meta.discourse.org/t/number-cpu-cores-incorrectly-detected-upcloud-com-vps/51950
here is fix for most architectures. There can still be some systems where cores are too slow and they should not be used as count for unicorn processes.


xfalcox:
LGTM!

What do you say @mpalmer ?

PR #308, opened 2016-11-11
launcher: silence which error when looking for docker/docker.io

alon:
One liner to silence the warning from which when it cannot find docker.io / docker. There is later a check for the docker_path being non empty, so the warning is not required and just noise for anyone (like me) without docker.io but with docker in their path.


SamSaffron:
so sorry that this has been left open for so long, I am confused cause `which` does not seem to be causing any errors here even when the file is not there, what errors are you seeing and how do I make them happen?

alon:
I don't recall, if I see it again I'll reopen.

PR #312, opened 2016-11-20
Add Travis and Coala for code linting

adhikasp:
Hello,

I see that this repo doesn't have any mean of CI nor linting. So I try to create one with Travis and [Coala](https://github.com/coala/coala) for for linting. It's a pretty big patch but mainly just editing of whitespace and newlines.

Here are list of rules that i use in this patch

## Markdown

- Set max line length 80
- Exception for line that have URL link
- Enforce `*` for bold and `_` for emphasis. I just adhere the most commonly used pattern previously.

## YAML

- Consistent indentation rules, based on same rules used in <www.yamllint.com>

## Ruby

- Enforce rules based on [RuboCop](https://github.com/bbatsov/rubocop)
- I disable few default rules because i dont have enough understanding in the codebase to make change without breaking things.

## Shell

- Actually i want to lint bash file too, but there are bugs in Coala that prevent excluding a rule that not relevant with this repo code. The rule is enforcing `cd .... || exit`, preventing script failing silently on changing dir.

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -1,0 +1,32 @@
+[default]
+bears = FilenameBear
+files = **.rb, **.md, **.yml
+ignore =
+
+[markdown]
+bears = MarkdownBear, LineLengthBear
+files = **.md
+language = markdown
+max_line_length = 80
+ignore_length_regex = .*(https?:\/\/) # ignore long link
+markdown_strong = *
+markdown_emphasis = _
+
+[ruby]
+bears = RubySyntaxBear, RuboCopBear
+files = **.rb
+language = ruby
+rubocop_config = .rubocop.yml
+
+# Disabled because ShellCheckBear enforce rule `cd ... || exit`
+# it's cannot disabled through coala.
+#[shell]
+#bears = ShellCheckBear
+#files = **.sh, image/base/*
+#shell = bash
+
+[yaml]
+bears = YAMLLintBear
+files = **.yml
+language = yaml
+yamllint_config = .yamllint
--- a/.coafile
+++ b/.coafile
@@ -1,6 +1,6 @@
 [default]
 bears = FilenameBear
-files = **.rb, **.md, **.yml
+files = **.md, **.yml
 ignore =
 
 [markdown]
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,7 +25,7 @@ Metrics/AbcSize:
 Performance/FixedSize:
     Enabled: false
 
-# Many line is tricky or became less unreadable when splited
+# Many line is tricky or became less readable when splited
 # especially line of executing shell command via run(...).
 Metrics/LineLength:
     Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,42 @@
+Stye/GlobalVars:
+    Enabled: false
+
+Style/Documentation:
+    Enabled: false
+
+# image/build.rb:8
+# image/build.rb:30
+# Methods have too many length
+Metrics/MethodLength:
+    Enabled: false
+
+# image/build.rb:16
+# Unhandled exception
+Lint/HandleExceptions:
+    Enabled: false
+
+# image/build.rb:30
+# Assignment Branch Condition size for build is too high. [22.29/15]
+Metrics/AbcSize:
+    Enabled: false
+
+# image/build.rb:32
+# Do not compute the size of statically sized objects.
+Performance/FixedSize:
+    Enabled: false
+
+# Many line is tricky or became less unreadable when splited
+# especially line of executing shell command via run(...).
+Metrics/LineLength:
+    Enabled: false
+
+# image/monitor/src/monitor.rb:24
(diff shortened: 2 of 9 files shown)


image/base/download_phantomjs
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-docker run --rm -it -v `pwd`:/tmp samsaffron/phantomjs cp /usr/bin/phantomjs /tmp
+docker run --rm -it -v '`pwd`:/tmp' samsaffron/phantomjs cp /usr/bin/phantomjs /tmp
```
jayvdb: this is semantically different, and incorrect.  `-v` argument is followed by '/HOST-DIR:/CONTAINER-DIR' , and shell needs to execute `pwd`, so it can not be in single quotes.

.yamllint
```diff
@@ -0,0 +1,10 @@
+
+rules:
+    # Too many long linux command that are tricky and become less readable
```
jayvdb: But, you have changed the line lengths of many YAML files.

I suggest that you undo those changes, as they are unnecessary for this patch, and they are not enforced by CI.

SamSaffron:
@adhikasp I totally appreciate the sentiment here, but feel that this change is huge and has a lot of risk of regression

adhikasp:
Update: 
- yml files now strictly just indentation consistency fix
- No bash file edit

@SamSaffron I could break the PR for just markdown file for now and maybe the YAML. Then in the future merge the Ruby. How about it?

jayvdb:
Is it possible to disable the indent rules for the Yaml ?

If only indentation rules are disabled, the Yaml changes should be minimal, and all of the other Yaml checks are then enforced to catch more important bugs in the future.

adhikasp:
@jayvdb Yes it's possible. The remaining fix in will be just few extraneous space and newline. The lint would still check for valid YAML file.

jayvdb:
IMO that would be better.  The yaml indentation style being used here is actually quite consistent and neat - so it isnt worth such a large amount of change to comply with the linter default indentation rules.

adhikasp:
FYI the indentation is **not** consistent. Example in [this line](https://github.com/discourse/discourse_docker/pull/312/files#diff-fefe86523a7ba4437965d353d971cd0fL39), the line 39 is not aligned with line 41, so I edit the line 41. I think about half of similliar portion of code like that is missing one space.

jayvdb:
Ya, IMO that is "quite" consistent, not not 100% consistent ;-)

adhikasp:
Update: I ignore the indentation problem according to @jayvdb idea, now the changed file is only markdown and ruby

.travis.yml
```diff
@@ -0,0 +1,3 @@
+services: docker
+script:
+  - docker run --volume=$(pwd):/app --workdir=/app coala/base:pre coala-ci --flush-cache
```
robbyoconnor: Line endings

.rubocop.yml
```diff
@@ -0,0 +1,42 @@
...
+# image/monitor/src/monitor.rb:24
+# Avoid using `rescue` in its modifier form.
+Style/RescueModifier:
+    Enabled: false
+
+# image/monitor/src/monitor.rb:82
+# Favor modifier `if` usage when having a single-line body.
+# ps: maybe a false positive
+Style/IfUnlessModifier:
+    Enabled: false
```
robbyoconnor: Line endings

robbyoconnor:
There shouldn't be carriage returns -- if you see ^M, change the line endings in your editor's config.

adhikasp:
@robbyoconnor Updated 👍

.coafile
```diff
@@ -0,0 +1,32 @@
+[default]
+bears = FilenameBear
+files = **.rb, **.md, **.yml
```
robbyoconnor: rubocop handles ruby for you.

.rubocop.yml
```diff
@@ -0,0 +1,42 @@
...
```
robbyoconnor: A lot of these shouldn't be disabled.

.travis.yml
```diff
@@ -0,0 +1,3 @@
+services: docker
+script:
+  - docker run --volume=$(pwd):/app --workdir=/app coala/base:pre coala-ci --flush-cache
```
robbyoconnor: Switch to `0.9` tag.

robbyoconnor:
@adhikasp see all my comments.

.rubocop.yml
adhikasp: Could you elaborate further on this? I'm not too familiar with the code base and ruby language yet so I'm afraid I will introduce regression bugs.

robbyoconnor: Therein lies the problem. If you're not familiar don't change things.

robbyoconnor:
@adhikasp Resolve conflicts.

SamSaffron:
I am happy to reopen some of this, but want to take baby steps here

first I 100% support just taking the rubocop config discourse project uses and applying it to and .rb files here.

PR #315, opened 2016-12-06
Some HTTPS improvements to achieve A+ on Qualsys SSL Labs

robbyoconnor:
[Make an A+ on Qualsys SSL Labs](https://www.ssllabs.com/ssltest/analyze.html?d=discourse.oconnor.ninja&s=104.236.222.245) -- It's always bugged me...figured it can't hurt to make a PR. 

- Make HSTS max-age longer for A+ on qualsys SSL labs
- dhparams 4096 bits vs 2048

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -88,5 +88,4 @@ hooks:
        filename: "/etc/nginx/conf.d/discourse.conf"
        from: /add_header.+/
        to: |
-         # remember the certificate for 80 days and automatically connect to HTTPS for this domain
-         add_header Strict-Transport-Security 'max-age=6912000';
+         add_header Strict-Transport-Security 'max-age=63072000';
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -3,7 +3,7 @@ run:
      cmd:
        # Generate strong Diffie-Hellman parameters
        - "mkdir -p /shared/ssl/"
-       - "[ -e /shared/ssl/dhparams.pem ] || openssl dhparam -out /shared/ssl/dhparams.pem 2048"
+       - "[ -e /shared/ssl/dhparams.pem ] || openssl dhparam -out /shared/ssl/dhparams.pem 4096"
   - replace:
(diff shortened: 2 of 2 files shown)


SamSaffron:
Looks good to me @mpalmer can your review and merge if you think it is good?

SamSaffron:
looks good!

robbyoconnor:
Thank you -- this scratched an itch I had!

coding-horror:
This mak3s a new Digital Ocean install take 40 minutes. Unacceptable and will be reverted.

xfalcox:
Reverted back to `2048`, since 4096 is [16x slower on average](http://security.stackexchange.com/questions/95178/diffie-hellman-parameters-still-calculating-after-24-hours)

```bash
time openssl dhparam -out /shared/ssl/dhparams.pem 2048
real    0m35.631s
user    0m35.352s
sys     0m0.068s

time openssl dhparam -out /shared/ssl/dhparams.pem 4096
real    48m55.420s
user    48m44.788s
sys     0m2.920s
```

robbyoconnor:
Yes, but having it higher means it's slightly more secure. You can generate the dhparams off of the server -- or even generate it while the container is running!

SamSaffron:
sure, but this PR impacts every new install. If there was a different pattern I would be more open to it. Still 40 minutes of CPU time is something you should opt in to.

robbyoconnor:
@SamSaffron there's a middle-ground -- dhparams can be generated while the container is running -- so we can get the system up and running; and THEN generate it.

SamSaffron:
how would NGINX even boot? also, you are going to have to come up with a PR that demos this technique

I am all for security but people need to opt-in to this level of pain.

robbyoconnor:
What if we make it configurable? I assume pups can pass parameters around right?

robbyoconnor:
The dhparams were the only thing that caused the discourse install to take longer -- everything else in this PR is benign,

SamSaffron:
of course, it should totally be configurable 

see the discussion here: https://meta.discourse.org/t/prime-calculation-for-openssl-just-goes-on/54653 

perhaps open a new discussion on meta about this and we can work something out.

PR #316, opened 2016-12-08
Memory Allocation Fiesta

xfalcox:
This PR:

- Changes default allocator to tcmalloc 2.5
- Makes possible to change to none, jemalloc 3.6 and latest lockless just by setting ALLOCATOR on the env key in the app.yml


SamSaffron:
would probably call it RUBY_ALLOCATOR except for that looks fine to me

PR #317, opened 2016-12-13
Adding dev dependencies in the docker image

ibnesayeed:
Adding `libsqlite3-dev` dependency for dev image while mailcatcher is added in the Gemfile of the main discourse repo.


SamSaffron:
thanks!

PR #319, opened 2016-12-25
FEATURE: Ask for default language in discourse-setup

caugner:
This solves the problem of Discourse generating default categories/topics in English rather than in the desired default language.

As suggested here: https://meta.discourse.org/t/default-categories-topics-in-english-despite-german-as-default-language-in-setup-wizard/54775/3?u=claas

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -219,6 +219,7 @@ ask_user_for_config() {
   local smtp_port="587"
   local smtp_user_name="postmaster@discourse.example.com"
   local smtp_password=""
+  local language="en"
   local letsencrypt_account_email="me@example.com"
   local letsencrypt_status="ENTER to skip"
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Don't we also want to set LANG while at it so it ensures Postgres uses the right locale for fulltext search ?

caugner:
> Don't we also want to set LANG while at it so it ensures Postgres uses the right locale for fulltext search ?

Is it [`emv.LANG`](https://github.com/discourse/discourse_docker/blob/e00fca6ef9e0f06e08705108fc26e7775dc19d3d/samples/standalone.yml) or [`params.db_default_text_search_config`](https://github.com/discourse/discourse_docker/blob/e00fca6ef9e0f06e08705108fc26e7775dc19d3d/samples/standalone.yml#L27) which controls the locale for fulltext search?

In general, we would need a mapping of languages to Postgres locales. Where would we maintain this and how do we make sure Discourse runs fine with these locales? (I think I remember having had upgrade issues when using `de_DE.UTF-8`, but I might be wrong.)

gschlager:
The list of currently supported locales can be found in https://github.com/discourse/translations-manager/blob/master/lib/translations_manager/locales.rb. That being said, having another list of supported locales in `discourse-setup` feels kinda wrong to me. I'd rather improve the setup wizard and the `default_locale` site setting, so that it updates seeded content accordingly. It's already on my todo list.

`default_text_search_config` could be removed from the Docker setup if we move that configuration into the app. Makes it a lot easier to map Discourse locales to PostgreSQL dictionaries. I think all the PostgreSQL methods related to full text search have an optional parameter for the dictionary.

As for the `LANG` environment variable. Looking at https://www.postgresql.org/docs/current/static/locale.html I don't think we need a lot of that functunality, if any. Maybe `LC_COLLATE` for sorting, but from the top of my head I can't name any columns where we actually need it. If we really need it, I'd say it probably better to configure collations on the column level instead of database level. That makes it a lot easier to reconfigure the collation when the `default_locale` is changed, since this could be done from the app. However, for that to work, we'd probably need to ship the Docker image with more locales.

caugner:
> I'd rather improve the setup wizard and the default_locale site setting, so that it updates seeded content accordingly.

Yes, that sounds much more elegant.

> It's already on my todo list.

Let's close this PR then, shouldn't we?

PR #324, opened 2016-12-29
Uses pg_ctl -w instead of postgres bin so we wait for pg readiness

xfalcox:


hyperknot:
Does wait have any effect here? Once it is run with background: yes, the other time it's run via runit, no waiting there either.

(Actually, can you explain why are there two different ways to start it? I mean line 40 and and 221 in master? Why not just start via runit?)

tgxworld:
@xfalcox Do we still need this?

xfalcox:
@tgxworld this only changes when your postgres is in recovery mode (because of an abrupt shutdown, for example).

Since this behavior isn't that common, I wasn't able to test it.

tgxworld:
@xfalcox I think this change is relatively safe. I think we can merge this if the commands are still valid for pg 9.5 :)

ZogStriP:
@xfalcox do we still need this?

PR #327, opened 2017-01-16
CoreOS Bug

Blackglade:
Accounts for the hostname CoreOs Bug. See here: https://github.com/coreos/bugs/issues/1764

### DIFF
--- a/launcher
+++ b/launcher
@@ -497,7 +497,14 @@ run_start() {
    done
 
    (
-     hostname=`hostname -s`
+     # To account for Docker Bug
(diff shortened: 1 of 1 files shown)


SamSaffron:
my concern is, who will remember to remove this when coreos fix the bug?

Blackglade:
you're right, we can just have upstream fix it in that case!

SamSaffron:
yeah, I would just hardcode hostname using env vars in the yml for now

See a few lines earlier.

PR #329, opened 2017-01-26
Discourse setup reconfig

pfaffman:
This PR allows people to re-run `./discourse-setup`. This is useful to:

- Turn on Let's Encrypt after initial testing (and DNS config) **tested**
- Turn off Let's Encrypt (but why?) **tested** (see note below)
- adjust `db_shared_buffers` and `UNICORN_WORKERS if droplet is resized  **tested**
- Change Email config **tested**
- Works with my install script unchanged

### Notes

- For some reason, this PR shows changes having to do with MacOS memory checking. I didn't do that and don't understand why it's showing up here; when I look at the current version that stuff is there. Apparently I still don't understand git.
- If Let's Encrypt is disabled, the templates are commented out, but `LETSENCRYPT_ACCOUNT_EMAIL` is left configured. This means that if the script is subsequently re-run, it'll default to re-enabling Let's Encrypt. I suppose I could change this, but it seems like an edge case.
- Defaults now come from `standalone.yml` rather than this script. Future changes to `standalone.yml` could break this script (but that's always been the case).
- I added a second email to `DISCOURSE_DEVELOPER_EMAILS`; before you had to look at the comments to see how to include two addresses (and the default in this script included two addresses).
- If there is a space between email addresses in `DISCOURSE_DEVELOPER_EMAILS` they are read incorrectly by `read_config()`. It's inconvenient, but a real live human will be seeing what's there, so this seems acceptable.
- I changed the first line to `#!/usr/bin/env bash`, which should let this script run on macs. Not sure why whoever wrote the mac memory code didn't do this.
- I changed the final `./launcher bootstrap $app_name` to `time ./launcher rebuild $app_name`. Rebuild works fine when there's no image to delete. I like seeing how long things take. I'll take out the `time` if it's deemed annoying.
- I added a `sleep 5` between when changes are made and when the rebuild starts to give people a chance to ^c before it starts. Seems minimally annoying, and it can be impossible to kill `bundle` with a ^C. I'll remove this if it's deemed troublesome.
- Old script refused to run if port 80 or 443 was occupied. Now there is also a message to `./launcher stop app` and then re-run.
- When I ran this on forum.literatecomputing.com I noticed that it fixed memory settings that I'd apparently not set correctly when I set it up by hand.
- None of the error checking code (`validate_config`) was changed, so it should be pretty safe.
- I think that I checked all the cases if a given variable is commented out (as most are in `standalone.yml`) or not (as when re-running the script).

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ##
 ## Make sure only root can run our script
@@ -27,13 +27,40 @@ check_and_install_docker () {
   fi
 }
 
+##
+## What are we running on
+##
+check_OS() {
+  echo `uname -s`
+}
+
+##
+## OS X available memory
+##
+check_osx_memory() {
+  echo `top -l 1 | awk '/PhysMem:/ {print $2}' | sed s/G//`
+}
+
+##
+## Linux available memory
+##
+check_linux_memory() {
+  echo `free -g --si | awk ' /Mem:/  {print $2} '`
+}
 
 ##
 ## Do we have enough memory and disk space for Discourse?
 ##
 check_disk_and_memory() {
 
-  avail_mem=`free -g --si | awk ' /Mem:/  {print $2} '`
+  os_type=$(check_OS)
+  avail_mem=0
+  if [ $os_type == "Darwin" ]; then
+    avail_mem=$(check_osx_memory)
+  else
+    avail_mem=$(check_linux_memory)
+  fi
+
   if [ "$avail_mem" -lt 1 ]; then
     echo "WARNING: Discourse requires 1GB RAM to run. This system does not appear"
     echo "to have sufficient memory."
@@ -43,11 +70,11 @@ check_disk_and_memory() {
     exit 1
   fi
 
-  if [ "$avail_mem" -lt 2 ]; then
+  if [ "$avail_mem" -le 2 ]; then
     total_swap=`free -g --si | awk ' /Swap:/  {print $2} '`
     if [ "$total_swap" -lt 2 ]; then
-      echo "WARNING: Discourse requires at least 2GB of swap when running with less "
-      echo "than 2GB of RAM. This system does not appear to have sufficient swap space."
+      echo "WARNING: Discourse requires at least 2GB of swap when running with 2GB of RAM"
+      echo "or less. This system does not appear to have sufficient swap space."
       echo
       echo "Without sufficient swap space, your site may not work properly, and future"
       echo "upgrades of Discourse may not complete successfully."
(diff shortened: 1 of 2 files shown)


xfalcox:
I didn't read the whole PR, but there's some white space changes too in `standalone.yml` that shouldn't be targeted at this PR, can you clean it please.

pfaffman:
I thought that my emacs settings were such that the Right Thing was happening and that, though whitespace-only commits were frowned upon, fixing them up on a commit was OK.

From https://meta.discourse.org/t/discourse-development-contribution-guidelines/3823

    no trailing whitespaces, blank lines should have no spaces

And

[quote]
Remember: We don't accept pull requests consisting entirely of style changes. Style changes in the context of pull requests that also refactor code, fix bugs, improve functionality are welcome.
[/quote]

It appears that when I saved, emacs removed spurious spaces on blank lines, which I thought was the desired behavior.

Do I understand correctly that you want me to go back and insert spaces on the five blank lines that erroneously had spaces on them before?

coding-horror:
Why are we adding OS X support? Why is that in scope? I don't think it is, and it complicates this PR.

coding-horror:
also why read the defaults from the standalone.yml? I'm more comfortable with them being hard coded here in this script.

ZogStriP:
This PR is too getting too big. @pfaffman, can you split it and do 1 PR per "changes"?

pfaffman:
## WTF, OS X?

Or, WTF, github?

The OS X support is in the current version. I don't understand how or why it looks like I submitted it. Could it be that I didn't pull down the current version before I branched? I guess I can try to start a new PR to solve this. I'm fairly flummoxed on this. Here's that OSX stuff in the current release.

Have a look:

https://github.com/discourse/discourse_docker/blob/master/discourse-setup#L56-L62

Presumably I can ditch this PR and make a new one that doesn't seem to blame me for code that's already there. I was as surprised as you to find this OS X stuff here, and it can't even work with the current `#!/bin/bash`.

## Why move defaults into standalone.yml?

My proposed script always reads app.yml to get its initial values, so there is only one way that it gets default values. That seems clean. If defaults for the first run of  `discourse-setup` live in `discourse-setup` and on subsequent runs, it loads them from `app.yml` then there are two code bases to maintain, one for initial runs of the script and another for re-runs of the script. If you prefer, I can have a branch that does one thing if app.yml already exists and another if it doesn't, but I think it'll be more complicated and more code. My proposed implementation also potentially lets `discourse-setup` work with other sample.yml files in the future.

coding-horror:
I see, fine on reading the file then but negatory on the OS X please

pfaffman:
Here's the thing on the OSX nonsense. I didn't do it. It is already in the current version. See?

https://github.com/discourse/discourse_docker/blob/master/discourse-setup#L56-L62

Here's a blame link. It was added 4 months ago.

https://github.com/discourse/discourse_docker/blame/master/discourse-setup#L37-L42

If you like, I can abandon this PR, do a fresh pull, create a new branch, and see i f that will make Github not show those changes in the PR. I don't know how it is that I am so inept at dealing with Git.

PR #331, opened 2017-01-30
Make our swap config compatible with systemd >= 207

xfalcox:


rene-gomez:
Hi team!
I have a come problems with this branch under jwilder/nginx-proxy, can you support me?

best regards

SamSaffron:
@xfalcox shouldn't this switch based on the OS? I am merging this now mostly cause I want this to hit the review site  and vast majority of people running discourse setup will be on systemd now

PR #332, opened 2017-01-31
Improve check for approved storage driver [Closes #310]

### LINKED ISSUE #310: docker-setup regex wrong
The regex which should filter out devicemapper, etc. does not work.

It's actually allowing almost everything, as `[aufs|btrfs|zfs|overlay]` actually matches a single character in the list `aufs|btrzovely`.

My idea would be to use something like this:
```
test=`$docker_path info 2> /dev/null | grep 'Storage Driver: '`
if [[ "$test" =~ "Storage\ Driver:\ (aufs|btrfs|zfs|overlay)" ]]
```

Somehow this didn't work, although regex101.com says it should work.

Without grep storage driver it'd output 3 lines:
```
Storage Driver: devicemapper
Logging Driver: json-file
Cgroup Driver: cgroupfs
```

mpalmer:
Previous test was both buggy and inefficient; we now avoid the need
to do a separate regex comparison.  I also improved the wording of
the error message somewhat to use Docker-Approved(TM) terminology.

### DIFF
--- a/launcher
+++ b/launcher
@@ -133,15 +133,15 @@ check_prereqs() {
     exit 1
   fi
 
-  # 2. running aufs or btrfs
-  test=`$docker_path info 2> /dev/null | grep 'Driver: '`
-  if [[ "$test" =~ [aufs|btrfs|zfs|overlay] ]] ; then : ; else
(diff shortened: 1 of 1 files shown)


SamSaffron:
looks good to me!

PR #333, opened 2017-02-08
Update HTTPS configuration

Angristan:
We should disable DH and use AES CBC/GCM + ECDH ciphers only. It's fast and secure, and widely compatible (cf https://tls.imirhil.fr/suite).
Also, HSTS should not be enabled by default, I think HSTS should be activated manually while knowing what are its consequences.

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -3,7 +3,6 @@ run:
      cmd:
        # Generate strong Diffie-Hellman parameters
        - "mkdir -p /shared/ssl/"
-       - "[ -e /shared/ssl/dhparams.pem ] || openssl dhparam -out /shared/ssl/dhparams.pem 2048"
   - replace:
      filename: "/etc/nginx/conf.d/discourse.conf"
      from: /server.+{/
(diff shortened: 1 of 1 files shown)


SamSaffron:
I strongly disagree with dropping HSTS by default, @mpalmer's call about changing anything with our cipher suite.

mpalmer:
Yeah, HSTS is where it's at.  Removing that significantly reduces security, and I'm not in favour of it.

As far as changing the cipher suites goes, what are the *actual* benefits of this change?  Is there one or more supported user agents which, with the current cipher suite set, select a sub-optimal cipher suite?  Also, I'll need to see what an ssllabs scan looks like for both the current and new configurations to decide whether whatever changes is acceptable.  If you want this change to happen before I get around to spinning up a scratch machine to point ssllabs at, you'll probably want to do those tests and paste the results here.

Angristan:
I'm okay with HSTS, I changed my mind.

I changed the cipher suite to show which ciphers are used.

With the new one we drop DH, 3DES, Camellia and ECDSA certificates because Let's Encrypt gives RSA certificates. We only use ECDH + RSA + AES 128/256 GCM/CBC + SHA/SHA256/SHA384.

3DES is weak, so we have to drop it. DH is ok, ECDH is faster and stronger, and does not require to generate a key. We use a secp384r1 ECDH key instead. There is no point to support Camellia because it does not add any compatibility.

Current configuration : 
screenshot_09-02-2017_b hwdf36cj

screenshot_09-02-2017_b rchv_dus

New configuration : 
screenshot_09-02-2017_b idnuuj7o

screenshot_09-02-2017_b m_otd-bv


According to SSLLabs, it removes support for : 
- IE 6 / XP
- IE 8 / XP
- Java 6u45 
- OpenSSL 0.9.8y
- Android 2.3.7

So only very old systems.

This PR is not a major security upgrade, just a little improvement. :)

mpalmer:
Thanks for the testing and details about compatibility.  Merged!

xfalcox:
This will make installs even faster, thanks @Angristan!

robbyoconnor:
I was _ALMOST_ objecting to this but nevermind :) Good work!

PR #336, opened 2017-02-24
Show how to grab the proper IP address behind a reverse proxy

Axel-Naumann:
Tell nginx to trust the ip address as provided by a forwarding reverse proxy.
Use X-Forwarded-For which is set by both Apache and nginx acting as reverse proxy.

This seems to be a common source of confusion, with hundreds of posts in meta.discourse,
including terrible "solutions" - so it might be worthwhile to clearly state the minimal functional solution.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -163,6 +163,16 @@ run:
       from: /client_max_body_size.+$/
       to: client_max_body_size $upload_size ;
 
+#  Uncomment the following section for discourse_docker behind a reverse proxy
+#  - replace:
+#      filename: "/etc/nginx/conf.d/discourse.conf"
+#      from: /listen 80;/
+#      to: |
+#         listen 80;
+#         
+#         set_real_ip_from <reverse-proxy-addr>; # address or CIDR
+#         real_ip_header X-Forwarded-For;
+
   - exec:
       cmd: echo "done configuring web"
       hook: web_config


karies:
For reference, these are some of the meta topics: https://meta.discourse.org/t/run-discourse-with-or-alongside-existing-apache-sites/19514 https://meta.discourse.org/t/whats-the-proper-way-to-set-up-discourse-behind-a-reverse-proxy/30262 https://meta.discourse.org/t/rate-limiting-when-behind-nginx-proxy/50947/12 

Especially the first one is a good motivation for a couple of commented lines ;-)

SamSaffron:
You can't have this in the web template, nobody looks there and nobody should edit it ever, at best it can be included in the application sample template. 

I worry that by even adding this to the application sample template you would cause a bunch of people to think about a problem they do not have.

karies:
LOL Fair enough. I'd have *loved* to have these lines somewhere "official" such that I can recognized this as a blessed solution (versus the other 50, like "just disable the throttler").

Can you think of a place to put this? Shall I submit a PR for the app sample template? (Or given the one-lineliness maybe you want to do it without PR?)

Let me know!

SamSaffron:
Sure, submit a PR with a one liner in `samples/standalone.yml` that should
do it.

atomaka:
> I worry that by even adding this to the application sample template you would cause a bunch of people to think about a problem they do not have.

I'm struggling to understand how this isn't a problem for everyone in some way. Using the discourse_docker (39805404) with `Docker version 17.06.0-ce, build 02c1d87`, my IP addresses were not being properly set, even after including `templates/cloudflare.template.yml`.  I had to trust the Docker host itself before Discourse started seeing the correct IP address:

```
set_real_ip_from 172.17.0.1;                                                                                                                                                          set_real_ip_from fe80::42:63ff:fe62:1d7a/64;
```

Am I missing something?

PR #344, opened 2017-04-06
Added environment variable for mail-receiver's smtp_should_reject support.

rcgordon:
These are the changes needed to the main Discourse container to fight backscatter. This isn't tested because I have no idea what I'm doing. :)

The discussion that led to this is here: https://meta.discourse.org/t/reducing-backscatter-in-email-interface/59974

There is also a PR for discourse/mail-receiver that relies on this PR: https://github.com/discourse/mail-receiver/pull/2

And the new API that mail-receiver will expect to call is on this PR: https://github.com/discourse/discourse/pull/4793

### DIFF
--- a/samples/mail-receiver.yml
+++ b/samples/mail-receiver.yml
@@ -26,6 +26,12 @@ env:
   ## the URL needs to have the subfolder included!
   DISCOURSE_MAIL_ENDPOINT: 'http://discourse.example.com/admin/email/handle_mail'
 
+  ## The URL of the smtp_should_reject endpoint of your Discourse forum.
+  ## This is your forum's base URL, with `/admin/email/smtp_should_reject.json`
+  ## appended.  Be careful if you're running a subfolder setup -- in that case,
+  ## the URL needs to have the subfolder included!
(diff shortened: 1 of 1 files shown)


mpalmer:
This looks fine to me, it'll just be blocking on landing the core and mail-receiver PRs.

rcgordon:
The core PR was merged (and Discourse is now alerting admins to update to v1.8.0.beta10, which includes that PR), so we just need to get this and the mail-receiver PR merged. Let me know if you need anything from me to get these accepted. Thanks!

rcgordon:
Just pinging to see if we can get these merged yet. Sorry to nag!

mpalmer:
An update on this: whilst preparing the updated containers, etc for this, I realised that we need to handle backwards compatibility somehow, otherwise people will need to update their existing `mail-receiver.yml` files alongside pulling the new container image, which is... optimistic, at best.  I'll have to put back some of the support for the old environment variable in the mail-receiver container before I can put this whole shenanigan live.  That'll be a little while, unfortunately.

PR #347, opened 2017-05-25
adding mysql dependency template to simplify usage of import scripts

lead4good:
[See also my post here](https://meta.discourse.org/t/how-to-use-the-bbpress-import-script-or-any-other-import-script-with-mysql-dependency/63349)


SamSaffron:
sure, this seems fine, thanks

PR #350, opened 2017-06-05
Update web.letsencrypt.ssl.template.yml

RoldanLT:
Improve SSL: 
- Extend HSTS config

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -89,4 +89,4 @@ hooks:
        filename: "/etc/nginx/conf.d/discourse.conf"
        from: /add_header.+/
        to: |
-         add_header Strict-Transport-Security 'max-age=63072000';
+         add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload';


SamSaffron:
doing preload takes a bunch of process, I am not sure I want to add it to the template. 

https://hstspreload.org/

mpalmer:
NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE NOPE

:wink: 

Never, ever, *EVER* put `preload` in a default configuration, anywhere, ever.  Ever.

EVAAAAAAAAAAH

There is a long and amusing list of people submitting bugs against Chromium saying "I turned off SSL because my ad provider didn't support HTTPS and now YOUR BROWSER SUCKS".  The reason always turns out to be that they copy-pasted a "secure ur site" example without knowing what they were doing, and it included `preload`, and some "helpful" soul submitted their site to the HSTS preload list, and it ended up in Chromium, and then they turned off HTTPS (sometimes they even did it properly and let the existing HSTS timeout expire!).

See also: https://hstspreload.org/#opt-in

I'm also not a fan of `includeSubDomains` in a template config like this, because, again, the chances of foot-pwnage by the unwary is way, way, *way* too high, to secure a situation that will very rarely occur.  Someone puts Discourse on their zone root, then tries to put something else that doesn't immediately do HTTPS on a subdomain, subtle breakage results.  In the common case, Discourse gets put on a terminal subdomain anyway, so there's no benefit to `includeSubDomains`.

I'd be OK with a PR that allowed people to easily enable these settings, if they so desired (preferably with some appropriately dire warnings, because I don't want too many people screaming on meta that we've destroyed their site), but as a default... well, see above.  :grinning:

PR #351, opened 2017-06-05
Update web.ssl.template.yml

RoldanLT:
Improve SSL: 
- Use [Mozilla SSL Intermediate Config](https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.11.6&openssl=1.0.2g&hsts=yes&profile=intermediate) 
- Enable Reuseport
- Extend HSTS
- Use proper return 301 instead of rewrite for HTTPS redirect

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -9,7 +9,7 @@ run:
      to: |
        server {
          listen 80;
-         rewrite ^ https://$$ENV_DISCOURSE_HOSTNAME$request_uri? permanent;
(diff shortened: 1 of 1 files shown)


SamSaffron:
Can we split this up into one PR per change, we went through the HTTPS stuff quite a few times, used to have a much longer cipher list

(sorry closing this cause the https stuff needs to be split off here) 

cc @mpalmer

RoldanLT:
I'll submit new PR without HTTPS.

PR #352, opened 2017-06-05
Update web.ratelimited.template.yml

RoldanLT:
Limit maximum connections per IP to combat HTTP/S Flood.
Normal browser doesn't allow more than 10 connections though :).


SamSaffron:
Looks good to me

PR #353, opened 2017-06-05
Update web.ssl.template.yml

RoldanLT:
Use proper return 301 instead of rewrite for HTTPS redirect

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -9,7 +9,7 @@ run:
      to: |
        server {
          listen 80;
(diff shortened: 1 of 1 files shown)


SamSaffron:
Can you explain why this is better? I would like to understand.

RoldanLT:
https://www.nginx.com/blog/creating-nginx-rewrite-rules/

The return directive is the simpler of the two general‑purpose directives and for that reason we recommend using it instead of rewrite when possible.

mpalmer:
I'm sold.  Happy to merge if you are, @SamSaffron.

PR #355, opened 2017-06-05
Update web.ssl.template.yml

RoldanLT:
Use [Mozilla SSL Intermediate Config](https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.11.6&openssl=1.0.2g&hsts=yes&profile=intermediate) ciphers suite.
 
For better browser compatability.


mpalmer:
I'm OK with adding chachapoly, but what browsers that Discourse supports does the rest of the existing config not cover?  That's a lot of extra ciphers, so what's the upside to them all?  Is there a shorter list that would produce an equivalent result?

RoldanLT:
I posted the compatability improvement here: https://meta.discourse.org/t/improve-ssl-use-mozilla-ssl-intermediate-config-enable-reuseport-extend-hsts/63927/2?u=roldanlt

On other sites I have I'm also using this cipher suite: https://github.com/cloudflare/sslconfig/blob/master/conf
But it needs patch to Nginx to fully work.

SamSaffron:
It is super unclear though how the compatibility improves with this change,
can you list the new devices/browsers that gain support after this change?

RoldanLT:
Compare this screenshot [here](https://meta.discourse.org/t/improve-ssl-use-mozilla-ssl-intermediate-config-enable-reuseport-extend-hsts/63927/2?u=roldanlt) vs [this one](https://www.ssllabs.com/ssltest/analyze.html?d=meta.discourse.org&s=64.71.168.222).

SamSaffron:
@RoldanLT its not fair to give people the gift of work 😸  

So you added support for IE8/XP and android 2.3.7 at the cost of making our cipher list huge. 

I am not seeing this as a win, IE8 can die a horrible death.

PR #366, opened 2017-08-09
add the :Z option to all mounted volumes so that discourse works when selinux is enabled

### LINKED ISSUE #365: launcher needs a small tweak before it works on a machine that has selinux enabled (Fedora, RHEL, etc)
The bind mounts just need the :Z option on them.

db48x:
fixes #365

### DIFF
--- a/launcher
+++ b/launcher
@@ -252,8 +252,12 @@ RUBY
 
 
 set_volumes() {
+  local volopts=
+  if selinuxenabled; then
+      volopts=":Z"
+  fi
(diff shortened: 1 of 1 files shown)


SamSaffron:
This looks reasonable, let's try it out!

SamSaffron:
Had to revert per: https://meta.discourse.org/t/update-to-1-9-0-beta5-fails-to-start-discourse-after-rebuild/67922/2

db48x:
hilariously, I just tripped over it myself. easy to fix, see the commit I just pushed.

SamSaffron:
yeah you we are going to need a new PR

PR #369, opened 2017-08-18
Add google chrome and NPM modules to discourse_test image

davidtaylorhq:
To allow qunit tests to be run using headless chrome. See https://meta.discourse.org/t/chrome-59-headless-support/62060/5?u=david_taylor


SamSaffron:
looks good to me, thanks!

PR #370, opened 2017-08-20
Bump node.js to version 8

davidtaylorhq:
FAO @SamSaffron following discussion here: https://meta.discourse.org/t/chrome-59-headless-support/62060/16?u=david_taylor

I have tested this by upgrading node to version 8, then running 
```
rake assets:precompile
```
in a standard production docker install. It completes without error, and the site appears to work fine.

Is there any further testing that should be done on this?


SamSaffron:
I support this, but may take a bit till we bump the image.

davidtaylorhq:
Great 😄 

I will wait until the base image is bumped before adding the `npm install` stuff to the dev/test images, as I see they are built more frequently than the base image.

PR #371, opened 2017-08-25
create discourse_test_multsite db in discourse_dev image

LeoMcA:


SamSaffron:
Thanks will kick off a build

PR #372, opened 2017-08-31
Building gems in parallels and use 4 workers

fantasticfears:


SamSaffron:
I am somewhat concerned here cause afaik we tried this in the past @tgxworld do you remember?

tgxworld:
@SamSaffron I can't remember if we tried this but on most basic digital ocean droplet it wouldn't make a difference since there is only one CPU. Despite that, the changes are pretty safe IMO.

fantasticfears:
It's a bit awkward for the worker number since `UNICORN_WORKERS` is twice as the number of cores. I don't know if I should use proc to detect that in the container.
And it's not going to help with complication of gem as far as I remember it's always set to maximum core number by Rubygems. They are a lot for Discourse. The chance is that might help with some IO tasks.
I tried 3 configs in basic DO which is 1.5GB, single core at fantasticfears-gsoc2017-01. Bundler version is 1.15.3. Here is the result.

```
time bundle install --deployment --verbose --without test --without development --retry 3 --jobs 4 --path vendor/bundle
real    10m30.159s
user    8m29.268s
sys     1m52.232s

time bundle install --deployment --verbose --without test --without development --retry 3 --jobs 2 --path vendor/bundle
real    11m37.074s
user    8m50.324s
sys     1m49.372s

time bundle install --deployment --verbose --without test --without development --retry 3 --jobs 1 --path vendor/bundle
real    11m43.260s
user    8m52.192s
sys     1m57.844s
```

tgxworld:
Interesting, one minute of savings :) I'm generally fine with this since I don't see too much of a risk here.

PR #377, opened 2017-10-08
Bootstrap should respect 'run_image' key from config (if present)

darwin:
I needed more fine-grained control over image name produced by launcher's bootstrap/rebuild tasks. I noted support for optional `run_image` config key. I believe this key should also affect bootstrap to produce image with the same name, rather than default hard-coded `local_discourse/<config>`.

Alternatives:

1) Or alternatively you could maybe consider adding a separate `build_image` config key. And I would make sure it matches my `run_image`.

2) Or to make it super flexible. Optional `image_name` should be introduced which should be used everywhere and can be optionally overridden by `run_image` or `build_image` for specific tasks (and for backwards config compatibility).

### DIFF
--- a/launcher
+++ b/launcher
@@ -567,6 +567,7 @@ run_bootstrap() {
 
   set_volumes
   set_links
+  set_run_image
 
   rm -f $cidbootstrap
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
I think we should add a build_image here, its a bit clearer. 

Then you would set build_image for bootstrap and run_image for running.

darwin:
Ok, thanks for response. I'm going to open a new PR.

PR #378, opened 2017-10-11
Fix Vagrantfile to run discourse successfully

hisashi-komine:
Fixed below:

1. Update box image from trusty64 to xenial64. (Xenial is suggested [here](https://github.com/discourse/discourse/blob/master/docs/INSTALL-cloud.md))
2. Install requirements on vagrant's shell provisioning.
3. Now won't create symlink from /vagrant to /var/discourse. Vagrant's synced_folder locks the files owner and makes the discourse installation fail.

Now vagrant-disksize is required to run `vagrant up` to expand disksize of vm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ image/docker-squash.tar.gz
 image/discourse_dev/postgres.template.yml
 image/discourse_dev/redis.template.yml
 .gc-state/*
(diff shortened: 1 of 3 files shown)


SamSaffron:
looks like this should work! thanks

PR #381, opened 2017-10-14
Set docker_min_version to 17.03.0

Flowdalic:
At the time of writing this, Amazon Linux AMI images come with docker
17.03. If we set it higher, then discourse instances running on those
images could not get *easily* updated.

Further motivation at https://github.com/discourse/discourse_docker/commit/ddf8c54cdde0e448526f4fa0e9f7272695db8271#commitcomment-24976766

### DIFF
--- a/launcher
+++ b/launcher
@@ -56,7 +56,7 @@ fi
 
 cd "$(dirname "$0")"
 
-docker_min_version='17.06.2'
+docker_min_version='17.03.0'
 docker_rec_version='17.06.2'
 git_min_version='1.8.0'
 git_rec_version='1.8.0'


SamSaffron:
Trouble is that 17.03 has a super nasty bug where storage explodes

Flowdalic:
> Trouble is that 17.03 has a super nasty bug where storage explodes

So nasty that it is better to have dozens of discourse installations which will not receive updates?

It's your choice, and I don't know the bug you are referring to, but wouldn't a warning message be the better option?

```bash
$ .\launcher rebuild app
WARNING: You are running a docker installation with a super nasty bug
causing the storage to explode. Update your docker installation as soon as possible!
```

jomaxro:
Minimum version was downgraded: https://github.com/discourse/discourse_docker/commit/0998a7c802006dfec5724c030131e7885c3e5d19

PR #385, opened 2017-12-12
Update samples

pfaffman:
Update `data.yml` and `web_only.yml` to match updated comments in `standalone.yml`.


SamSaffron:
thanks!

PR #386, opened 2017-12-18
Added support for unattended deployment

erseco:
Added support for unattended deployment reading environment variables for the different setup options, an example of calling:

```
sudo HOSTNAME=discourse.example.com DEVELOPER_EMAILS=me@example.com SMTP_ADDRESS=smtp.example.com SMTP_PORT=587 SMTP_USER_NAME=user@example.com SMTP_PASSWORD=pa$$word LETSENCRYPT_ACCOUNT_EMAIL=me@example.com AUTO_START=y ./discourse-setup
```

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -325,41 +325,58 @@ ask_user_for_config() {
 
   while [[ "$config_ok" == "n" ]]
   do
-    if [ ! -z $hostname ]
-    then
-      read -p "Hostname for your Discourse? [$hostname]: " new_value
-      if [ ! -z $new_value ]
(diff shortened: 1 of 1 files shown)


SamSaffron:
I support this change @mpalmer does it look good to you?

mpalmer:
I like the concept, but the current implementation isn't up to scratch.  We're using `bash`, there's no need for the `"X${VAR}" = "X"` test (`-z "$VAR"` works just fine).  Also, I think it's well past time the "ask a question, get a variable value" pattern in there got moved into a function -- pass it an env var and a prompt, get back an answer.  That'd simplify the PR immensely.

PR #387, opened 2017-12-28
Update samples

pfaffman:
update data.yml and web_only.yml to match what's in standalone.yml. (Also, add an undocumented "2container" switch to discourse-setup that will install separate data and web containers.)

Sorry for the multiple commits. Hopefully it's not too hard to squash them on your end.


SamSaffron:
OK lets try this out, cleanup alone is worth it here.

PR #388, opened 2018-01-01
Fix URL to Docker base image

dnsmichi:


SamSaffron:
hmmm @xfalcox  is this correct post your refactor of images

dnsmichi:
Was fixed with #390.

PR #389, opened 2018-01-14
For resumed installation fix config file backup.

fwolfst:
When installation is aborted and later "resumed" (setup
script called a second (third, ...) ) time, following error will
show
"""
The configuration file containers/app.yml already exists!

. . . reconfiguring . . .

Saving old file as app.yml.2018-01-14-232141.bak
cp: missing destination file operand after
'containers/app.yml.2018-01-14-232141.bak'
Try 'cp --help' for more information.
"""
I believe this is due to using the wrong variable ($config_file vs
$web_file), which is fixed with this commit.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -629,7 +629,7 @@ then
   DATE=`date +"%Y-%m-%d-%H%M%S"`
   BACKUP=$app_name.yml.$DATE.bak
   echo Saving old file as $BACKUP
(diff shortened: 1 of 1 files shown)


xfalcox:
Looks like this was fixed on https://github.com/discourse/discourse_docker/commit/275735af30b9f158dc41097683b450ab478cbc4b

PR #394, opened 2018-06-01
FIX: disk space check tests /shared

pfaffman:
If `/shared` is not the same as `/`then the disk space test fails because it's testing the wrong partition.
For example, if `/` is too small you might create a separate partition for postgres so there will be enough room to update.


tgxworld:
Oops sorry we missed this. Thanks @pfaffman

pfaffman:
Cool. No problem for me. I have not yet needed it again yet. :-)

If there's something I can do to make PRs more obvious or whatever, please let me know.  I'm still figuring this out.

Cheers.

PR #395, opened 2018-06-20
Make launcher work with french locale (among others)

mscherer:
On several locales others than C, A-Z might match more than A to Z, due
to different collation order.

See https://unix.stackexchange.com/questions/227070/why-does-a-z-match-lowercase-letters-in-bash
for explanation.


dgutride:
This change would be very valuable for us - can it be reviewed?

SamSaffron:
OK we can try this I guess, thanks!

PR #396, opened 2018-07-07
FEATURE: Add the doctor

pfaffman:
At long last!

### DIFF
--- a/discourse-doctor
+++ b/discourse-doctor
@@ -68,7 +68,7 @@ check_ip_match() {
 }
 
 check_docker_is_installed() {
-  log -e "\n\n==================== DOCKER INFO ===================="
+  log -e "\n==================== DOCKER INFO ===================="
   docker_path="$(which docker.io || which docker)"
   if [ -z $docker_path ]; then
     log "Docker is not installed. Have you installed Discourse at all?"
@@ -77,12 +77,12 @@ check_docker_is_installed() {
     exit
   else
     log -e "DOCKER VERSION: $(docker --version)"
-    log -e "DOCKER PROCESSES\n\n $(sudo docker ps -a)\n"
+    log -e "\nDOCKER PROCESSES (docker ps -a)\n\n$(sudo docker ps -a)\n"
   fi
 }
 
 get_OS() {
-  "OS: $(uname -s)\n"
+  log -e "OS: $(uname -s)"
 }
 
 check_disk_and_memory() {
@@ -95,20 +95,20 @@ check_disk_and_memory() {
   fi
   log "$(free -m)"
 
-  log -e "\n\n==================== DISK SPACE CHECK ===================="
-  log System Disk Space
+  log -e "\n==================== DISK SPACE CHECK ===================="
+  log "---------- OS Disk Space ----------"
   log "$(df -h / /var/discourse /var/lib/docker /var/lib/docker/* | uniq)"
 
   if [ "$version" != "NOT FOUND" ]
   then
     log
-    log Container Disk Space
+    log "---------- Container Disk Space ----------"
     log "$(sudo docker exec -w /var/www/discourse -i $app_name df -h / /shared/ /shared/postgres_data /shared/redis_data /shared/backups /var/log | uniq)"
   fi
 
-  log -e "\n\n==================== DISK INFORMATION ===================="
+  log -e "\n==================== DISK INFORMATION ===================="
   log "$( fdisk -l )"
-  log -e "\n\n==================== END DISK INFORMATION ====================\n\n"
+  log -e "\n==================== END DISK INFORMATION ===================="
 
   free_disk="$(df /var | tail -n 1 | awk '{print $4}')"
   # Arguably ./launcher is doing this so discourse-doctor does not need to . . .
@@ -159,7 +159,7 @@ check_if_hostname_resolves_here() {
 ## get discourse configuration values from YML file
 ##
 get_discourse_config() {
-  log -e "\n\n==================== YML SETTINGS ===================="
+  log -e "\n==================== YML SETTINGS ===================="
   read_config "DISCOURSE_HOSTNAME"
   DISCOURSE_HOSTNAME=$read_config_result
   log DISCOURSE_HOSTNAME=$DISCOURSE_HOSTNAME
@@ -185,7 +185,7 @@ get_discourse_config() {
 
 check_plugins() {
   log -e "\n\n==================== PLUGINS ===================="
-  log -e $(grep git containers/$app_name.yml)
+  log -e "$(grep git containers/$app_name.yml)"
   grep git containers/$app_name.yml > /tmp/$PPID.grep
 
   if grep -cv "github.com/discourse" /tmp/$PPID.grep > /dev/null
@@ -194,9 +194,9 @@ check_plugins() {
     log You have what appear to be non-official plugins.
     log "If you are having trouble, you should disable them and try rebuilding again."
   else
-    log "No non-official plugins detected"
+    log -e "\nNo non-official plugins detected."
   fi
-  log
+  log -e "\nSee https://github.com/discourse/discourse/blob/master/lib/plugin/metadata.rb for the official list.\n"
 }
 
 dump_yaml() {
@@ -218,9 +218,9 @@ read_config() {
 ## call rake emails:test inside the container
 ##
 check_email() {
-  log -e "\n\n==================== MAIL TEST ===================="
+  log -e "\n==================== MAIL TEST ===================="
   log "For a robust test, get an address from http://www.mail-tester.com/"
-  log "Or just send a test message to yourself."
+  echo "Or just send a test message to yourself."
   EMAIL=$(echo $DEVELOPER_EMAILS |cut -d , -f 1)
   read -p "Email address for mail test? ('n' to skip) [$EMAIL]: " new_value
   if [ ! -z "$new_value" ]
@@ -267,14 +267,14 @@ check_docker() {
     log "$app_name not running!"
     log "Attempting to rebuild"
     log "==================== REBUILD LOG ===================="
-    log "$(./launcher rebuild $app_name)"
+    # too hard to pass STDERR of ./launcher to log()
+    ./launcher rebuild $app_name 2>&1 | tee -a $LOG_FILE
     log "==================== END REBUILD LOG ===================="
     docker ps| tail -n +2 > /tmp/$UUID-docker.txt
     if grep $app_name /tmp/$UUID-docker.txt
     then
-      log -e "\n\n==================== REBUILD SUCCEEDED! ====================\n\n"
-      log "Discourse container $app_name is running!"
-      log "Waiting 30 seconds for container to crank up. . . "
+      log -e "\nDiscourse container $app_name is now running."
+      log ". . . waiting 30 seconds for container to crank up. . . "
       sleep 30
     else
       log "$app_name still not running!"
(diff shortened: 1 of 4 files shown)


discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+      echo "You should probably remove any non-standard plugins and rebuild."
+      NO_CONTAINER='y'
+    fi
+  fi
+}
+
+##
+## Discourse version
+##
+get_discourse_version () {
```
jjaffeux: there's `get_discourse_version()` and `discourse_version()` could we make the difference more explicit please?

discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+## What are we running on
+##
+check_OS() {
+  echo "----------------------------------------"
+  echo -n "OS:"  |tee -a $LOG_FILE
+  echo `uname -s` |tee -a  $LOG_FILE
+}
+
+##
+## OS X available memory
```
jjaffeux: I would prefer if we didn’t have comments which are  basically the same than the method name please.

discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+  read_config_result=`echo $config_line | awk  --field-separator=":" '{print $2}'`
+  read_config_result=`echo $read_config_result | sed "s/^\([\"']\)\(.*\)\1\$/\2/g"`
+}
+
+##
+## send a test message
+##
+email_test() {
+  echo -e "\n\n==================== MAIL TEST ===================="
+  echo "I recommend that you get a test address from http://www.mail-tester.com/ , but you "
```
jjaffeux: Can we phrase it in a more general way please? eg: avoid the "I" :D

discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+check_OS() {
+  echo "----------------------------------------"
+  echo -n "OS:"  |tee -a $LOG_FILE
+  echo `uname -s` |tee -a  $LOG_FILE
+}
+
+##
+## OS X available memory
+##
+check_osx_memory() {
```
jjaffeux: AFAIK discourse-doctor is for supported docker installs and discourse officially supports only 64 bit Linux compatible with Docker . SO I'm unsure if we should bother with macOS stuff here.

discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+
+##
+## Do we have enough memory and disk space for Discourse?
+##
+check_disk_and_memory() {
+
+  os_type=$(check_OS)
+  avail_mem=0
+  if [ "$os_type" == "Darwin" ]; then
+    avail_mem=$(check_osx_memory)
```
jjaffeux: where is avail_mem used ?

discourse-doctor
```diff
@@ -0,0 +1,403 @@
...
+
+}
+
+##
+## output YML file
+##
+dump_yaml() {
+  echo -e "\n\n==================== YML DUMP ====================" | tee -a $LOG_FILE
+  echo Dumping $app_name.yml | tee -a $LOG_FILE
+  sed -e '/\s*#.*$/d' -e '/^\s*$/d' -e 's/  DISCOURSE_SMTP_PASSWORD:\(.*\)/  DISCOURSE_SMTP_PASSWORD: REDACTED/g' containers/$app_name.yml  | tee -a $LOG_FILE
```
jjaffeux: are we not already doing this in `clean_up_log_file()` ?

discourse-doctor
pfaffman: This should be cleaned up. Sorry.

discourse-doctor
pfaffman: That's code I pulled from discourse-setup, & someone else wrote it. I agree that they are superfluous.

discourse-doctor
pfaffman: So noted.

"For a robust test, get an address from..." ?

discourse-doctor
pfaffman: I totally agree, and I don't know how that OSX code snuck in to `discourse-setup`, but tons of people (for reasons that I don't understand) try to run Discourse on a Mac, and regularly seek help, so we may as well leave it in.

discourse-doctor
pfaffman: So rather than calling clean_up_log_file once, you want to call it every time the file is appended? That would mean, for example, piping `./launcher rebuild app`, though the cleanup script, which seems fraught.

Oh, I see. When I started, I didn't realize that I'd need to redact from the output of the rebuild, so I thought I could just catch it as I went in to the file.

I removed the piecemeal redactions and have just the one call to `clean_up_log_file` at the end.

discourse-doctor
pfaffman: In `discourse-setup`. I pulled out the variable.

discourse-doctor
pfaffman: Yeah. Those aren't great names. How about  `get_discourse_version` that `wget`s `$VERSION_HOSTNAME` and `compare_discourse_versions` that checks whether `localhost` and `$DISCOURSE_HOSTNAME` match?

I'll remove the comments too.

Also, `check_IP_match` (also pulled from `./discourse-setup`) wasn't getting called. I tweaked those comments and added a call to `check_IP_match`.

discourse-doctor
pfaffman: I pulled out all of the ones that were near-duplicates of the function name. Again, I am pretty sure that someone added these to my original PR of `discourse-setup`  & I inferred that it was Discourse Style. If everyone agrees that such comments are superfluous and inappropriate, I should remove them from `discourse-setup` too. (And then I wonder if the code that's used by both `discourse-setup` and `discourse-doctor` should be pulled in to a library that gets sourced by both of them. . . )

pfaffman:
I think I've done all that you asked. I'll see if it still works and re-submit Real Soon Now.

pfaffman:
Hey, @jjaffeux  and @davidtaylorhq. Here's another shot at this.

Thanks.

discourse-doctor
```diff
@@ -0,0 +1,374 @@
...
+  else
+    echo "NO K"
+    return 1
+  fi
+}
+
+check_ip_match () {
+  HOST="$1"
+  echo
+  echo Checking your domain name . . . | tee -a $LOG_FILE
```
jjaffeux: Could we have generic logging function?

So we would have
```
log("Checking your domain name ...")
```

instead of

```
echo Checking your domain name . . . | tee -a $LOG_FILE
```

This would remove a lot of redundant code.

discourse-doctor
```diff
@@ -0,0 +1,374 @@
...
+check_linux_memory() {
+  echo "----------------------------------------"
+  echo -n "RAM (MB):" |tee -a $LOG_FILE
+  free -m --si | awk ' /Mem:/  {print $2} ' | tee -a $LOG_FILE
+  free -m | tee -a $LOG_FILE
+}
+
+check_disk_and_memory() {
+
+  os_type=$(check_OS)
```
jjaffeux: so we don’t print anything if memory is low? we just print the available memory ?

discourse-doctor
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $DIR
+
+LOG_FILE="/tmp/discourse-debug.txt"
+rm -f $LOG_FILE
+echo DISCOURSE DOCTOR `date` | tee -a $LOG_FILE
```
jjaffeux: I'm not bash script expert but I think it's better to write:

```
echo "DISCOURSE DOCTOR $(date)" | tee -a $LOG_FILE
```

discourse-doctor
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $DIR
```
jjaffeux: I think `cd $DIR || exit` would be better

discourse-doctor
```diff
@@ -0,0 +1,374 @@
...
+  if [ "$version" != "NOT FOUND" ]
+  then
+    echo | tee -a $LOG_FILE
+    echo Container Disk Space | tee -a $LOG_FILE
+    sudo docker exec -w /var/www/discourse -i $app_name df -h / /shared/ /shared/postgres_data /shared/redis_data /shared/backups /var/log|uniq | tee -a  $LOG_FILE
+  fi
+
+
+  echo -e "\n\n==================== DISK INFORMATION ====================" >> $LOG_FILE
+  sudo fdisk -l >> $LOG_FILE
```
jjaffeux: This might be problematic https://github.com/koalaman/shellcheck/wiki/SC2024

discourse-doctor
```diff
@@ -0,0 +1,374 @@
...
+##
+
+docker_check() {
+  if [ -f /var/discourse/containers/app.yml ]
+  then
+    echo "YML=app.yml"
+    app_name=app
+    data_name=app
+    web_file=containers/$app_name.yml
+    data_file=containers/$app_name.yml
```
jjaffeux: where is this variable `data_file` being used ?

discourse-doctor
```diff
@@ -0,0 +1,374 @@
...
+## get discourse configuration values from YML file
+##
+
+get_discourse_config() {
+
+  echo -e "\n\n==================== YML SETTINGS ====================" | tee -a $LOG_FILE
+  read_config "DISCOURSE_HOSTNAME"
+  DISCOURSE_HOSTNAME=$read_config_result
+  echo DISCOURSE_HOSTNAME=$DISCOURSE_HOSTNAME  | tee -a  $LOG_FILE
+  local changelog=/tmp/changelog.$PPID
```
jjaffeux: where is this used?

discourse-doctor
pfaffman: Well, my way was better 25 years ago, but it looks like you're right today. :-)

https://stackoverflow.com/questions/9405478/command-substitution-backticks-or-dollar-sign-paren-enclosed

discourse-doctor
pfaffman: Except then the script would quit without explanation. I guess that would happen only if they try to run the script without first changing to it and the directory where `discourse-doctor` lives did not have the X bit set for that user.

I'll stick bit after the root check, so it should never fail.

discourse-doctor
pfaffman: No. It can't be problematic, because they have to run the script as root, but it is superfluous.

discourse-doctor
pfaffman: It's not, but if we add a Postgres version check in the future, we have a hook to be able to know how to get into the data container in the event that they have a separate data container that's called data.

If you want, I'll remove it.

discourse-doctor
pfaffman: In `discourse-setup`. I'll pull it out.

discourse-doctor
jjaffeux: It's up to you to write an explanation here before exiting. But I think it's a good practice to ensure cd was a success otherwise you could end up continuing the script thinking you are in a specific dir when you are not. And if today it makes sense, and nothing is risky after, who knows what will be put after in the future, that's why I recommended going with the safe choice here.

pfaffman:
It's cleaned up now and I think it does what we want.

davidtaylorhq:
Looks great to me - I experimented with it on a standard installation and tried breaking a few standard things (stop docker container, third party plugins, incorrect DNS). All visible in the doctor's output :)

PR #398, opened 2018-07-21
Added "exec" option to Launcher

ryanerwin:
https://meta.discourse.org/t/running-commands-inside-discourse-container/92811

Allows running CLI commands directly from the host
For example:
./launcher exec app discourse enable_restore
./launcher exec app discourse restore $file
to run a restore on a staging server via cron script

### DIFF
--- a/launcher
+++ b/launcher
@@ -8,6 +8,7 @@ usage () {
   echo "    restart:    Restart a container"
   echo "    destroy:    Stop and remove a container"
   echo "    enter:      Open a shell to run commands inside the container"
+  echo "    exec:       Execute command line inside the container"
   echo "    logs:       View the Docker logs for a container"
   echo "    bootstrap:  Bootstrap a container for the config based on a template"
   echo "    rebuild:    Rebuild a container (destroy old, bootstrap, start new)"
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am so sorry, I just don't feel right adding this to launcher. docker exec does the trick anyway

pihentagy:
> I am so sorry, I just don't feel right adding this to launcher. docker exec does the trick anyway

Maybe but isn't it a hack rather than a "trick"?

Please add the wasted hours figuring this "trick"  with a few hours.

How are the two below should be trivially equivalent for a discourse administrator who just want to automate restoring?

```
~/infra/discourse/discourse on  master ⌚ 23:15:29
$ vagrant ssh -c 'cd /vagrant;sudo ./launcher enter discourse'
root@ubuntu-xenial-discourse:/var/www/discourse# discourse enable_restore
Restore are now permitted. Disable them with `disable_restore`
root@ubuntu-xenial-discourse:/var/www/discourse# 
```
and
```
vagrant ssh -c 'sudo docker exec -w /var/www/discourse -i discourse discourse enable_restore'
```

PR #399, opened 2018-07-23
SECURITY: Add HSTS directives includeSubDomains and preload.

MarianoRD:
To better comply with the HSTS directive, improve SSL security and add the possibility of all Disocurse sites with HTTPS to be registered in the Chrome HSTS preload list.

Also remove what I think is a unneeded ````add_header```` in ````templates/web.ssl.template.yml````


SamSaffron:
I think both of these should be opt in, preload requires extra work and includesubdomain is only relevant if you have subdomains

xfalcox:
Closed as those aren't generally applicable for everyone.

PR #405, opened 2018-08-28
Remove unneeded "restricted" Ubuntu repository

sudoman:
This change removes the `restricted` Ubuntu repository from the base image, because it could install non-free software. It is unneeded, as no non-free software is installed by the base Docker file.

The Docker image successfully compiled without any errors from `apt`. Note that I tried also tried to remove the `universe` repo, but it is currently needed for the `fping` package, so this commit leaves that untouched.

I ran the automated tests for this change, but they timed out early, possibly due to slow private key generation steps.

[Discussion on meta.discourse.org](https://meta.discourse.org/t/docker-image-includes-unneeded-ubuntu-restricted-non-free-repo/88931?u=sudoman)


xfalcox:
This was finally done on https://github.com/discourse/discourse_docker/commit/223b69e775c61a7fb05386262281f7fa2f0e9520

PR #406, opened 2018-09-01
ruby-china change domain

scavin:


SamSaffron:
@xrav3nz is this change correct?

xrav3nz:
LGTM 👍 just tested locally as well

PR #409, opened 2018-09-28
Use rm -rf in remove-old-socket

khannurien:
In some cases, the leftover socket is actually a directory.


SamSaffron:
This is fine with me! thanks!

PR #410, opened 2018-10-03
Run Rake task to fetch MaxMindDb.

nbianca:
To be merged after discourse/discourse#6335.

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -139,4 +139,5 @@ RUN useradd discourse -s /bin/bash -m -U &&\
     cd /var/www/discourse &&\
     sudo -u discourse bundle install --deployment \
          --without test --without development &&\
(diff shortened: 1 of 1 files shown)


gschlager:
Hmm... I think this will break for users who use the current Docker image with an older Discourse version (e.g. stable). We probably need to test if the rake task actually exists before we run it.

SamSaffron:
nahh this should be totally safe, its simply in the base image which is always built with tests-passed

ZogStriP:
Hard to revert this since using `rake` loads `rails` which loads the site settings, which requires a connection to Redis and 💥

SamSaffron:
> Hard to revert this since using rake loads rails which loads the site settings, which requires a connection to Redis and 

Hmmm ... we should fix it ups then so this rake task loads less stuff... having a quick look

PR #411, opened 2018-10-04
Move tmp directories for backups and restores into /shared

gschlager:
Large temporary files shouldn't be stored inside of the Docker container.


SamSaffron:
This looks good to me, having this data in AUFS layers is not good, nice catch

PR #412, opened 2018-10-09
download MadMindDB

ZogStriP:
Here's my take to download the MaxMindDB in the base image.

@SamSaffron what do you think?


SamSaffron:
OK sorted this out by fixing the rake task in discourse not to require environment

PR #413, opened 2018-11-07
Configure the nginx resolver to eliminate name resolution on startup

mwcampbell:
This is desirable for two reasons:

1. As reported in a [bug topic](https://meta.discourse.org/t/restarting-container-when-internet-is-down-causes-discourse-to-fail/42615/4), if nginx resolves Internet hostnames on startup, then the container won't run if the host is disconnected from the Internet at startup.

2. If the IP address of the avatars.discourse.org hostname ever changes, then nginx will end p using an old IP address until the administrator restarts the container, unless we use dynamic resolution.

This commit adds a simple script to extract the nameserver IP addresses from `/etc/resolv.conf` and write them to an nginx configuration file. The configuration change for Discourse itself will be in a separate PR in the main Discourse repository.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -56,6 +56,15 @@ run:
         [ ! -z "$COMPRESS_BROTLI" ] && sed -i "s/. brotli/  brotli/" /etc/nginx/conf.d/discourse.conf || sed -i "s/. brotli/# brotli/" /etc/nginx/conf.d/discourse.conf
 
   - file:
+     path: /etc/runit/1.d/configure-nginx-resolver
+     chmod: "+x"
+     contents: |
+        #!/bin/bash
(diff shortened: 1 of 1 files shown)


SamSaffron:
I have not forgotten about this it just that the change feels very high risk to me, what if the grep here somehow generates an invalid file in some conditions and then nginx has trouble dealing with it.

I am not against this, its just that I worry a lot about impact especially since the fix itself is so rarely needed.

mwcampbell:
Fair enough. I could attempt a more "proper" resolv.conf parser, or look for existing code to reuse. But it may not be worth it.

SamSaffron:
so sorry, we ended up doing the proxy in the app so this change is no longer needed.

PR #414, opened 2018-12-20
Launcher script breaks at high numbers of sites

eduardopoleo:
On a multi-site app configuration files such as `containers/web.yml` grow with the number of sites. We discovered that the launcher script breaks at around 100 sites due to "Argument List too long" when commands such `echo` are being referenced from the within an `exec` call.

This workaround echoes the `$input` content outside the exec (which we verified works fine for 300 sites) and then cats the file content inside the exec. Thus bypassing the argument limitation. Finally, we remove the temp_input file generated.

### DIFF
--- a/launcher
+++ b/launcher
@@ -335,7 +335,12 @@ set_template_info() {
     puts env.map{|k,v| "-e\n#{k}=#{v}" }.join("\n")
 RUBY
 
-    raw=`exec echo "$input" | $docker_path run $user_args --rm -i -a stdin -a stdout $image ruby -e "$env_ruby"`
+    tmp_input_file=$(mktemp)
+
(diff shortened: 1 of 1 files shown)


SamSaffron:
this looks a bit strange to me ... temp input belongs in `TMPDIR` no?

eduardopoleo:
@SamSaffron updated. We are now using `mktemp` which creates the file on the `TMPDIR` if set otherwise uses `/tmp`

SamSaffron:
sorry this is taking so long, looks like it leaves around a temp file in certain cases ... maybe make the tempfile just before you need it and then remove once done.

SamSaffron:
OK, this looks reasonably safe, lets try it! thanks

PR #415, opened 2018-12-25
Add hacks to make it work on aarch64

luqasn:


luqasn:
Sorry, misclicked! not production ready :)

PR #418, opened 2019-01-28
Configure Git in test image

techAPJ:
This should fix the failing specs for `discourse-code-review` repo.


image/discourse_bench/Dockerfile
```diff
@@ -7,10 +7,6 @@ FROM discourse/discourse_test:1.4.0
 # Install ApacheBench
 RUN apt-get install -y apache2-utils libsqlite3-dev
 
-# configure Git to suppress warnings
```
techAPJ: Removed from here since `discourse_bench` image builds on top of `discourse_test`.

SamSaffron:
looks good to me!

PR #419, opened 2019-02-11
merge docker_args with user_args for launcher file

deargle:
see https://meta.discourse.org/t/starting-discourse-on-a-specific-docker-network/49401/4

I took a best guess and added a merging function call to the end of set_template_info because it looked like that was called at the start of a function whenever something config-file-related was used.

### DIFF
--- a/launcher
+++ b/launcher
@@ -626,7 +626,7 @@ run_start() {
 
      set -x
      $docker_path run --shm-size=512m $links $attach_on_run $restart_policy "${env[@]}" "${labels[@]}" -h "$hostname" \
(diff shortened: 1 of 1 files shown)


deargle:
Shoot hold on got mixed up on a merge and undid a recent `expose` change, brb

SamSaffron:
This does looks safe to me @saj  can you give it a quick review ?

saj:
[Merged](https://github.com/discourse/discourse_docker/commit/bfc79e773024fbbc80af3253b7a5aaa904a26079) with one slight change.

Thank you for the patch.

PR #416, opened 2019-01-04
Use latest docker garbage collector

sau226:
This patch configures discourse_docker to use the latest docker_gc version as the version we use was last updated 3 years ago and (in my opinion) a bit slow compared with this version.

PR #420, opened 2019-02-13
Use docker system prune

sau226:
This commit updates the repository to use Docker's built in method of
cleaning up old containers per Sam Saffron's request in #416. 

The cleanup command and 1 instance where `docker system prune` was already used will call `docker system prune -f` (don't prompt the user for a yes/no to trigger the command) as the launcher script will prompt the user for confirmation before running `docker system prune`.

P.S: If you are working for Discourse and have write access to this repo feel free to add the `-a` (remove stopped containers as well as all unused images) option to where I've used `docker system prune`. I want your opinon on this so if you want/need it then feel free to commit that to my fork before merging (I'm fine with that).

### DIFF
--- a/launcher
+++ b/launcher
@@ -226,7 +226,7 @@ check_prereqs() {
     echo
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
-      docker system prune
+      docker system prune -f
       echo "If the cleanup was successful, you may try again now"
     fi
     exit 1
@@ -468,7 +468,7 @@ fi
       space=$(df /var/lib/docker | awk '{ print $4 }' | grep -v Available)
       echo "Starting Cleanup (bytes free $space)"
 
-      STATE_DIR=./.gc-state scripts/docker-gc
+      docker system prune -f
 
       rm -f shared/standalone/log/var-log/*.txt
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Thanks heaps for the feedback here, I went ahead with the change (see last 2 commits) I strangely forgot to push my first attempt at it yesterday. 

Let me know what you think.

sau226:
@SamSaffron I'd suggest using restoring the old code (in ./launcher only) and restoring the code from your first commit (but replace `STATE_DIR=./.gc-state scripts/docker-gc` with `$docker_path system prune -a -f` or similar).

Being a developer I (and probably at least 1 other person outside your team that uses Discourse) want the ability to see how much data is being saved. If that isn't a strong enough justification Discourse also purges logs by running `rm -f shared/standalone/log/var-log/*.txt` after the gc script would otherwise run.

If needed I can file a PR for this or an issue on meta. Let me know what you think.

SamSaffron:
not sure about this `*.txt` in var logs is a relic of the past, no need to worry about cleaning it up anymore.

if we can figure out extra stuff to remove on cleanup I am open to that ... but I fine the text we get back from `prune -a` to be fine for now.

PR #421, opened 2019-02-19
fix: ampersand with no quotes truncates echo output

massimog:
This just gives a wrong indication to user


SamSaffron:
goops, thanks for the fix!

PR #425, opened 2019-03-18
Bump acme.sh to 2.8.0

sau226:
Bump the acme.sh script (Let's Encrypt) version for more API options, ACME v2 wildcard certs and bug fixes

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -10,7 +10,7 @@ hooks:
 
     - exec:
        cmd:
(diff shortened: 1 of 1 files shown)


SamSaffron:
@tgxworld  what are your thoughts here? 

@sau226 what are the advantages of the upgrade?

tgxworld:
Acme.sh actually has automatic upgrade built in if I recall correctly so this probably won’t make a difference more most instances.

sau226:
Actually it looks like we (Discourse) manually force a specific commit hash so this commit updates the hash. I can modify the PR to just use the latest version provided if you want.

tgxworld:
@sau226 This looks OK. Thank you for contribution :+1:

PR #427, opened 2019-04-25
Update ImageMagick to 7.0.8-42

neemiasvf:
This version of ImageMagick (7.0.8-41) is not available anymore. Accessing https://imagemagick.org/download/ImageMagick-7.0.8-41.tar.gz will return a 404. So, building a Docker image with that version will return an error as well.

### DIFF
--- a/image/base/install-imagemagick
+++ b/image/base/install-imagemagick
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -o errexit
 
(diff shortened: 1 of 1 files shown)


gschlager:
I'd rather switch back to using the GitHub mirror than update ImageMagick every week or so. Please hold tight a couple of days... I'd like to fix it with https://github.com/discourse/discourse_docker/pull/429

PR #428, opened 2019-04-25
Use HEAD instead of @ shortcut for git in launcher

greedy:
The abbreviation @ for HEAD was added in git 1.8.5. The launcher claims to be compatible with git version 1.8.0 but the use of this abbreviation breaks this compatibility. This change is needed to support RHEL 7.6 which has only git 1.8.3.1.

### DIFF
--- a/launcher
+++ b/launcher
@@ -760,9 +760,9 @@ case "$command" in
 
         git remote update
 
-        LOCAL=$(git rev-parse @)
+        LOCAL=$(git rev-parse HEAD)
         REMOTE=$(git rev-parse @{u})
-        BASE=$(git merge-base @ @{u})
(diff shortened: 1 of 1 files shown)


SamSaffron:
I prefer to just drop support for git 1.8... it is way ancient.

greedy:
Sure, it's old but it seems unfortunate to insist RHEL users install a new git just to make the up-to-date check for the launcher script work. Especially since the fix is so simple.

SamSaffron:
I guess, it is super ancient though. I will merge this, but I worry that other issues are headed your way especially if you are not running overlay

greedy:
That’s the thing about RHEL. It has super old stuff like this in the userland but the kernel has the necessary backports to use overlay2.

PR #429, opened 2019-04-25
Update dependencies

gschlager:
* Ruby 2.6.3 which has a couple of Unicode improvements

* nginx from 1.5.9 to 1.5.12 (http://nginx.org/en/CHANGES)

* ImageMagick 7.0.8-42 and switch it back to using GitHub, because only the latest release is available on the official site and this regularly breaks our build

* libpng from 1.6.36 to 1.6.37 (security fix)

* gifsicle from 1.91 to 1.92 (http://www.lcdf.org/gifsicle/changes.html)

* Node.js v10, the latest active LTS (https://nodejs.org/en/about/releases/)


SamSaffron:
Nice, all the changes look good.

PR #431, opened 2019-06-13
discourse_test image: include official plugins and their gems

pmusaraj:
To speed up plugin ruby and js test builds, it's useful to have the official plugins and their gems already in the test image.


SamSaffron:
Looks good, go ahead and merge!

PR #432, opened 2019-06-14
Repo key should be downloaded securely

maticmeznar:
PostgreSQL repository public key should be downloaded using a secure connection, otherwise anyone performing MITM attack can do bad things.


SamSaffron:
absolutely! nice catch.

PR #437, opened 2019-07-13
Update web.ssl.template.yml - drop cbc & use X25519 curve

mcmufffin:
X25519 see: https://safecurves.cr.yp.to/
drop cbc see: https://blog.qualys.com/technology/2019/04/22/zombie-poodle-and-goldendoodle-vulnerabilities


xfalcox:
Thanks for the PR @mcmufffin !

We just updated our SSL cypher list to match https://github.com/mozilla/server-side-tls/tree/last-revision-before-move, which I think will cover our needs.

PR #439, opened 2019-08-14
Fixup: use latest image discourse/base:release

lord2y:
Instead of using an old outdate image this script should use the latest.


SamSaffron:
we are holding off here for a week or so. you can override on local if you wish.

PR #444, opened 2019-09-07
FEATURE: Elliptic Curve certificate

gschlager:
[Mozilla](https://wiki.mozilla.org/Security/Server_Side_TLS) recommends ECDSA (P-256) as certificate type for intermediate compatibility.

> ECDSA certificates are recommended over RSA certificates, as they allow the use of ECDHE with Windows 7 clients using Internet Explorer 11

Most modern browsers will use cipher suites with the ECDSA certificate. Older browsers will select the RSA certificate and a RSA cipher suite.

It will create two Let's Encrypt certificates:
* EC 256 bits (SHA256withRSA) 
* RSA 4096 bits (SHA256withRSA)

Without this change all the ECDSA cipher suites defined in https://github.com/discourse/discourse_docker/blob/12f501764f57c827e497eb6fb88e98f8c3c468e6/templates/web.ssl.template.yml#L22 won't work. With the new certificate all cipher suites will work and browsers like IE11 on Windows 7 and Windows 8 will work too.

**Before:**

**After:**

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -10,7 +10,7 @@ hooks:
 
     - exec:
        cmd:
-         - cd /root && git clone https://github.com/Neilpang/acme.sh.git && cd /root/acme.sh && git reset --hard f62a4a0c0ccf6cd73b5746dd8b8790ce3c512833
+         - cd /root && git clone --branch 2.8.2 --depth 1 https://github.com/Neilpang/acme.sh.git && cd /root/acme.sh
          - touch /var/spool/cron/crontabs/root
          - install -d -m 0755 -g root -o root $LETSENCRYPT_DIR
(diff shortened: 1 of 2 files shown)


SamSaffron:
@mpalmer  as our resident SSL guru, thoughts?

mpalmer:
Do we care about supporting IE11 on Windows 7?  There's no security downside to presenting two certificates, as far as I'm aware, other than the code complexity and handshake size.  Could probably mitigate the handshake size by dropping back to a 2048 bit RSA key (it really is overkill to be generating a 4k RSA key for a 90 day cert).

SamSaffron:
Unfortunately we need to wait till June 2020 to drop IE11, that said Windows 7 support is gone in Jan 2020. 

I am fine to drop down to a 2048 bit RSA key if you would like to follow it up Gerhard. Or we can just wait till June next year and simplify our ciphers.

gschlager:
The ECDSA certificate actually reduces the handshake size, because the server sends only 1 certificate after the client announced its capabilities. I guess lowering to 2048 RSA keys is fine as well. I can follow up. Most modern browsers will select the ECDSA key anyway.

PR #447, opened 2019-09-18
FIX: Never prune Docker volumes

saj:
`system prune` on older Docker releases will remove volumes.  The
accidental removal of container volumes may result in user data loss.

This patch should ensure that any users on Docker CE <17.06.1 benefit
from the same, safer behaviour enjoyed by users on contemporary Docker
releases.


SamSaffron:
Feel free to merge away

PR #448, opened 2019-09-19
Make port check optional, add y/n prompt

RubenHoms:
Based on [discussion on forums](https://meta.discourse.org/t/discourse-setup-assumes-netcat-is-installed/128893/2), this will improve the flow for installation. It will check if netcat is installed, and if it isn't present the user with an option; continue without port check or abort to install `netcat` and rerun the setup.


SamSaffron:
This looks reasonably safe to me, lets give it a shot!

PR #451, opened 2019-10-04
DEV: Bump uglifyjs

romanrizzi:


SamSaffron:
I am super happy with this but we are going to need to do a careful coordinated dance here. I guess the thing to confirm here is that 

1. Core fix is cross compatible (works with bot 3 and 2) 
2. Backport the uglify 3 support to stable / beta 
3. Bump the image 

Cause we don't want to break stable here. 

Once you are ready with all the lego pieces feel free to merge.

romanrizzi:
> Core fix is cross compatible (works with bot 3 and 2) 
✅ 

> Backport the uglify 3 support to stable / beta

I backported it to stable, do we need to cut a new beta to include it there?

SamSaffron:
We should probably cut a new beta anyway with the all the recent fixed @nlalonde

PR #452, opened 2019-10-24
Change check for linux memory

recursivecodes:
Some VMs clock in at *just under* 1GB, so checking for 1GB of RAM will miss these.  Instead, check for MB, divide by 1000 and round up.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -114,7 +114,8 @@ check_osx_memory() {
 ## Linux available memory
 ##
 check_linux_memory() {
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am not sure about this change, I can accept 990MB being ~ 1GB, but 501MB is a stretch.

recursivecodes:
> I am not sure about this change, I can accept 990MB being ~ 1GB, but 501MB is a stretch.

Fair point. I was trying to keep it a one liner, but I've just updated it to be more precise.

SamSaffron:
I am fine with this change now, thanks!

PR #454, opened 2020-01-19
Update README.md

AnonaWolf:


jomaxro:
@AnonaWolf, what is the intent of this change? It seems to only add a line break.

CvX:
Looks like a spam PR, see also: https://github.com/Homebrew/formulae.brew.sh/pull/222. Closing.

PR #457, opened 2020-02-28
FIX: mysql-dep template--un-freeze the Gemfile

pfaffman:
I suspect that the others of the import templates need this as well.


gschlager:
Thanks! Fixed in https://github.com/discourse/discourse_docker/commit/bb9a17362619770e5ae0f416c87d321414c8d0c7

PR #458, opened 2020-03-08
Don't use fully qualified path for bash in host

mikroskeem:
My main reason for that is NixOS not having bash at /bin/bash. Besides, shebang line already uses `/usr/bin/env` to find bash location, thus this also makes script re-execing more consistent 😄

### DIFF
--- a/launcher
+++ b/launcher
@@ -788,7 +788,7 @@ case "$command" in
           do
             args[$j]=${BASH_ARGV[$i]}
           done
-          exec /bin/bash $0 "${args[@]}" # $@ is empty, because of shift at the beginning. Use BASH_ARGV instead.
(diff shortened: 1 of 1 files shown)


mikroskeem:
On a second thought, `bash` argument could even be removed from `exec`, assuming that `launcher` is always executable and `$0` points to the actual script file reliably.

ZogStriP:
Is that the only spot where we're using `bash`?

mikroskeem:
If I didn't miss something, then that was the only place where bash was executed with absolute path (I'm not near computer right now to triple check). Other occurances of `/bin/bash` were used inside Discourse container, which is okay.

SamSaffron:
cool, I think this should work, bash should certainly be in PATH

PR #463, opened 2020-05-06
FIX: Add a blank sshd.template.yml

riking:
Fixes https://meta.discourse.org/t/cant-rebuild-because-ssh-template-is-removed/150285/10


Supermathie:
I had pondered this very thing… perhaps we should instead add a template that says "Remove the sshd template from your app.yml and rebuild"

riking:
Sure, I'll merge this and send another because I suspect we'll have some debate about the method there...

PR #465, opened 2020-05-16
Ensure it overrides all locale related variables

yaegashi:
`LC_ALL`, `LANG` and `LANGUAGE` are set to `en_US.UTF-8` in the docker image.  The config should override all of them when other locale is needed.  Otherwise it fails to create a DB with the proper locale setting.

I actually failed to upgrade my DB from Postgres 10 to 12 with `launcher rebuild app`:

```console
Building PostgreSQL dictionaries from installed myspell/hunspell packages...
Removing obsolete dictionary files:
Stopping PostgreSQL 10 database server: main.
Stopping PostgreSQL 12 database server: main.
Performing Consistency Checks
-----------------------------
Checking cluster versions                                   ok
Checking database user is the install user                  ok
Checking database connection settings                       ok
Checking for prepared transactions                          ok
Checking for reg* data types in user tables                 ok
Checking for contrib/isn with bigint-passing mismatch       ok
Checking for tables WITH OIDS                               ok
Checking for invalid "sql_identifier" user columns          ok
Creating dump of global objects                             ok
Creating dump of database schemas
  discourse
  postgres
  template1
                                                            ok

lc_collate values for database "postgres" do not match:  old "ja_JP.UTF-8", new "en_US.UTF-8"
Failure, exiting
```

You might want to have the launcher to auto correct the inconsistency of the locale variables.

### DIFF
--- a/samples/data.yml
+++ b/samples/data.yml
@@ -21,7 +21,9 @@ params:
 
 env:
   # ensure locale exists in container, you may need to install it
+  LC_ALL: en_US.UTF-8
   LANG: en_US.UTF-8
+  LANGUAGE: en_US.UTF-8
 
 volumes:
   - volume:
--- a/samples/mail-receiver.yml
+++ b/samples/mail-receiver.yml
@@ -14,7 +14,9 @@ expose:
   - "25:25"   # SMTP
 
 env:
+  LC_ALL: en_US.UTF-8
   LANG: en_US.UTF-8
+  LANGUAGE: en_US.UTF-8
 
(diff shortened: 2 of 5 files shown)


SamSaffron:
I do support this change @xfalcox does it look good?

xfalcox:
> I do support this change @xfalcox does it look good?

Looks good, but as it only updates samples it won't fix current installs. I will also add some notes to our update manual.

Thanks @yaegashi !

xfalcox (approved):

PR #466, opened 2020-05-22
Fix install suffix of jemalloc

dLobatog:
Just a nitpick, I found that the version of jemalloc-new installed (5.2.0) doesn't correspond to the version in the suffix.


SamSaffron:
Thanks heaps... I just updated to 5.2.1 so we are on latest stable for "new" jemalloc

Still not seeing much perf benefit but we will re-test.

PR #467, opened 2020-05-24
launcher: Separate "update" command for updating git repository

tormodvolden:
When simply doing a "rebuild" it would also update the git
repository, which is not always desirable.

It is better to amend instructions to:
./launcher update
./launcher rebuild
for situations where the previous behaviour is wanted.

Signed-off-by: Tormod Volden <debian.tormod@gmail.com>
---

Alternatively we could just rip out this git update stuff, and amend instructions to:
git pull
./launcher rebuild
which is also more transparent.

Otherwise it would be necessary to update the documentation (also the inline one) to point out that "rebuild" will update the git repository.

### DIFF
--- a/launcher
+++ b/launcher
@@ -11,6 +11,7 @@ usage () {
   echo "    logs:        View the Docker logs for a container"
   echo "    bootstrap:   Bootstrap a container for the config based on a template"
   echo "    run:         Run the given command with the config in the context of the last bootstrapped image"
+  echo "    update       Update this git repository (recommended before rebuild)"
   echo "    rebuild:     Rebuild a container (destroy old, bootstrap, start new)"
   echo "    cleanup:     Remove all containers that have stopped for > 24 hours"
   echo "    start-cmd:   Generate docker command used to start container"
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am sorry, but I disagree with this quite strongly. People will forget to update and rebuild is already out there as our recommendation for getting latest version.

You already have a mechanism to skip pre reqs, why not use that?

tormodvolden:
Fair enough if you have been documenting this everywhere as the way to update everything in one command. IMO poorly named from the start though, "rebuild" should be building up again what you have, not pulling in new unknowns. People using git as a revision tool are not used to having the git tree change under their feet while running programs in the repo.

Anyway, I would come back to my fallback suggestion, better explain the "rebuild", especially in the inline help that currently doesn't mention updating the repo.

The --skip-prereq option does not change this behaviour. The git update is inside the "rebuild" block.

One could think of a --skip-update option, but the way the script works today, running destroy / bootstrap / start would do the same, you just have to know it.

PR #468, opened 2020-05-25
FIX: Automatically retry image pull once

riking:
This works around the DockerHub TLS errors.


SamSaffron:
K ... lets try it 

I wonder if we should also sleep 1 between the 2 tries.

PR #474, opened 2020-07-14
Build libheif from source for Ubuntu

pmusaraj:
Followup to https://github.com/discourse/discourse_docker/commit/8dc661f5a1a4c971db5fc7e01826a12f984fe56d, we use Ubuntu in CI and Ubuntu requires `libheif` to be built from source.


xfalcox:
Shouldn't this remove the lib installed in https://github.com/discourse/discourse_docker/commit/8dc661f5a1a4c971db5fc7e01826a12f984fe56d#diff-43b0d5b4b3f55138e81153c35c8ecb30R17 so we don't have two of the same polluting the path?

Also, maybe let's install libaom so we get AVIF for free?

pmusaraj:
@xfalcox I tried adding `libaom-dev` but it causes ImageMagick to not include support for `heic`. Will leave it out for now.

pmusaraj:
Merged by hand in https://github.com/discourse/discourse_docker/commit/c53ee530a7013de126580ccf7420486f8b8c98a3

PR #476, opened 2020-07-27
Add Support for Btrfs Storage Driver

mritzmann:
We are running the btrfs storage driver for a while on a discourse forum with 4,3k users, and everything is working fine. Since we have to adjust the `launcher` every time, i submit this small merge request.

### DIFF
--- a/launcher
+++ b/launcher
@@ -168,7 +168,7 @@ check_prereqs() {
   fi
 
   # 2. running an approved storage driver?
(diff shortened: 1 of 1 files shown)


SamSaffron:
I am afraid we can not merge this, you will have to skip pre-reqs there is a flag for it you can use.

See: https://meta.discourse.org/t/unsupported-storage-and-checking-latest-launcher-create-infinite-loop/117014 for context

mritzmann:
I don't see any specific reason for this, but I'm sure you will have your reasons.

PR #478, opened 2020-08-04
FIX: discourse-doctor plugin check too loose

pfaffman:
The non-official plugin check was looking only for `git`, which would match other things in the yml file (e.g., a digital ocean S3 bucket name).

`grep` for `'git clone'` should solve that problem.

Though I did test this edit on a running sitee, I made these edit in the web interface by hand rather than pushing a commit where I tested it. Please look twice to make sure that I didn't do something silly when making the edit here.


SamSaffron:
thanks yeah seems reasonable.

PR #480, opened 2020-08-10
Handle the case where IPv6 is also used

Supermathie:
* We want web.ssl.template.yml to handle the IPv6 case as well

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -12,11 +12,24 @@ run:
        }
        server {
   - replace:
-     hook: ssl
      filename: "/etc/nginx/conf.d/discourse.conf"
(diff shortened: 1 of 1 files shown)


Supermathie:
Don't merge this yet, I'm still testing it, but this makes certain workflows a bit nicer

jomaxro:
You can make it a draft if not yet safe to merge.

Supermathie:
another thing I'd like to consider is that this `sed` feels brittle - should we change the discourse_docker base image to:
* have a marker for replacement
* listen on both v4 and v6 by default

SamSaffron (approved):

SamSaffron:
I think this is fine, but why did you remove the hook?

Supermathie:
I *moved* the hook.

PR #481, opened 2020-08-21
FEATURE: Add easy mechanism for skipping post-deployment migrations

riking:
This will help self-hosters avoid longer downtime when a large post-deployment migration is pending.

Usage:

```
./launcher rebuild app --skip-post-migrations
./launcher enter app
SKIP_POST_DEPLOYMENT_MIGRATIONS=0 rake db:migrate
exit
./launcher restart app
```

(Note: the 'restart' clears the temporarily-set environment variable, so future web upgrades will correctly apply post-deployment migrations.)

https://meta.discourse.org/t/2-6-0b2-upgrade-very-slow/161604/19?u=riking

### DIFF
--- a/launcher
+++ b/launcher
@@ -20,6 +20,7 @@ usage () {
   echo "    --docker-args              Extra arguments to pass when running docker"
   echo "    --skip-mac-address         Don't assign a mac address"
   echo "    --run-image                Override the image used for running the container"
+  echo "    --skip-post-migrations     Apply the SKIP_POST_DEPLOYMENT_MIGRATIONS=1 environment variable"
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


SamSaffron:
Not a huge fan of this, it is adding more complexity. 

Just add `SKIP_POST_DEPLOYMENT_MIGRATIONS` to ENV and rebuild ... then zero it out in your exec.

riking:
The idea here is that this avoids two extra trips through your editor, and prevents a lot of mistakes from people forgetting to remove the env var afterwards.

SamSaffron:
I am sorry @riking closing this. 

Launcher is just so enormous, I want it to have less options and be less bash. This feature is just a bit too rare to use.

PR #483, opened 2020-10-16
Replace the default ImageMagick policy.xml

jbrw:
Replace the default ImageMagick policy.xml with a file containing some resource limitiations. The intention is to stop exceptionally oversized images (and/or malformed images) from consuming all resources on a system.

The values provided should provide ample resources for any reasonable image to be processed without hinderance.


jbrw:
I'm not super-familiar with discourse_docker, so there could well be much better ways to do this.

Any feedback is greatly appreciated!

SamSaffron:
I think it probably makes sense just to add this to the web template, if we add another template here then everyone are going to need to add it to their config which will likely be forgotten.

SamSaffron:
This looks good to me. Lets try it out!

SamSaffron (approved):

PR #484, opened 2020-11-04
FEATURE: Run themes:update before assets:precompile

udan11:
Task was implemented in discourse/discourse#11102.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -169,7 +169,8 @@ run:
       cd: $home
       hook: assets_precompile
       cmd:
-        - su discourse -c 'bundle exec rake assets:precompile'
+        - su discourse -c 'bundle exec rake themes:update assets:precompile'
+
   - file:
      path: /usr/local/bin/discourse
      chmod: +x


ZogStriP (approved):

SamSaffron:
I worry here that we are spinning a whole new process and that this will add another big delay to builds. 

Can we do this as part of asset precompile (at the end) ... if some sort of flag is specified maybe? 

Also docker manager will miss this if we do this here.

udan11:
How would you feel about combining the invocation of `themes:update` with `plugin:pull_compatible_all`?

Instead of:

```
bundle exec rake plugin:pull_compatible_all
bundle exec rake themes:update
```

There would be only:


```
bundle exec rake plugin:pull_compatible_all themes:update
```

SamSaffron:
pull_compatible_all though does not even boot rails, so it will be pretty fast. this trick will not actually save much time.

I think some special ENV we pass to precompile is probably best approach, that way we can do all the updates upfront just before precompile using the same booted instance of rails.

udan11:
Ah, I missed that Rails is not booted for `plugin:pull_compatible_all`. I do not like the `ENV` solution because it needs extra code to handle that and I think Rake should handle this just fine.

I would change:

```
bundle exec rake assets:precompile
```

to

```
bundle exec rake themes:update assets:precompile
```

This boots Rails only once and runs both tasks.

SamSaffron:
This is close... but what we need to minimize booting is: 

`DISCOURSE_UPDATE_THEMES=1 bundle exec rake assets:precompile'

This will mean we only boot rails once for this process, keep in mind on digital ocean: 

```
/# time rails r puts


real	0m5.739s
user	0m4.532s
sys	0m0.964s
```

So that is 5.7 seconds just to boot rails to do nothing, so this change would add that 5-10 seconds everywhere unless we pass it in using env.

udan11:
I am adding a clarification here. Rake will only boot Rails once even if there is more than a task requiring environment.

Let's consider two tasks, both booting Rails:

```
task "foo" => :environment do
  puts "foo"
end

task "bar" => :environment do
  puts "bar"
end
```

Wee see that there is no difference between executing foo+bar or just foo or bar.

```
$ time bin/rake foo bar
foo
bar
bin/rake foo bar  2.69s user 0.97s system 97% cpu 3.772 total

$ time bin/rake foo
foo
bin/rake foo  2.69s user 0.97s system 98% cpu 3.725 total

$ time bin/rake bar
bar
bin/rake bar  2.72s user 1.00s system 96% cpu 3.863 total
```

xfalcox:
This change breaks rebuilding stable.

SamSaffron:
we can add a blank themes:update to stable and beta @udan11

PR #485, opened 2020-11-04
Remove additional exit command

haomingw:
The `usage` function contains already the `exit 1`


SamSaffron:
thanks! looks good to me.

PR #489, opened 2020-11-23
Adding sed statement to disable imklog in rsyslog

limed:
It appears the rsyslog version in `debian:buster-slim` has the `imklog` module syntax as follows

```
module(load="imklog")   # provides kernel logging support
```

This PR comments the `imklog` portion out, I'm leaving the previous method of commenting out the line to make sure it doesn't break anything. However happy to remove it if its deemed necessary


SamSaffron:
This seems like a nice change, less annoying non-error messages. 

Thanks!

PR #492, opened 2020-12-01
launcher: add argument for user-specified base image.

fitzy101:
This is useful for testing changes to the base image used for the app
container.

### DIFF
--- a/launcher
+++ b/launcher
@@ -20,6 +20,7 @@ usage () {
   echo "    --docker-args              Extra arguments to pass when running docker"
   echo "    --skip-mac-address         Don't assign a mac address"
   echo "    --run-image                Override the image used for running the container"
+  echo "    --base-image               Override the base image used for the container"
   exit 1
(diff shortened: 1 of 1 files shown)


xfalcox:
The base_image can already be set in the app.yml file, or using a different default in the launcher script on a branch. Not sure about carrying this. What is your use case?

SamSaffron (approved):

fitzy101:
> The base_image can already be set in the app.yml file, or using a different default in the launcher script on a branch. Not sure about carrying this. What is your use case?

Use case is for easily testing changes for the base image on any branch - I can see now there is an override that can be declared in the `app.yml` once that has been generated. I was changing the image manually in the launcher script but found it cumbersome. @xfalcox if you'd prefer to just keep the existing override I'll close this PR.

fitzy101:
Closing in favour of using existing mechanisms.

PR #493, opened 2020-12-02
Update base image and default to postgres 13.

fitzy101:
This updates the base image to Postgres 13, as well as including a new template for postgres 13.

### DIFF
--- a/templates/postgres.template.yml
+++ b/templates/postgres.template.yml
@@ -67,7 +67,7 @@ run:
           echo "WARNING: Upgrading PostgresSQL would require an addtional $(numfmt --to=si $(($required - $free_disk))) of disk space"
           echo "Please free up some space, or expand your disk, before continuing."
           echo ''
-          echo 'To avoid upgrading change "templates/postgres.template.yml" TO "templates/postgres.10.template.yml" in containers/app.yml'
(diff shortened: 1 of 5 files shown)


templates/postgres.template.yml
```diff
@@ -167,36 +168,36 @@ run:
...
+      filename: "/etc/postgresql/13/main/pg_hba.conf"
       from: /^#local +replication +postgres +peer$/
       to: "local replication postgres  peer"
 
   # allow all to connect in with md5 auth
   - replace:
-      filename: "/etc/postgresql/12/main/pg_hba.conf"
-      from: /^host.*all.*all.*127.*$/
+      filename: "/etc/postgresql/13/main/pg_hba.conf"
+      from: /^host.*all.*all.*137.*$/
```
CvX, line 193: Search-and-replace-all strikes again! 😄

I believe it should remain unchanged?

```suggestion
      from: /^host.*all.*all.*127.*$/
```

fitzy101: I thought I got all these! Good spotting.

xfalcox:
Looking good!

fitzy101:
@xfalcox I've pushed a few commits since Friday, and reverted the change to the image as that will need to change once this lands on master and the new image is built and tagged. I'll follow up with another commit to update that once the image is built. Would you mind reviewing again? I've also tested in a variety of disk usage scenarios, and have a Meta post ready to go once things are merged.

PR #498, opened 2020-12-09
Remove `host_run` from `launcher`.

tgxworld:
This is legacy code that we no longer use and support.


jomaxro (approved):

tgxworld:
@SamSaffron Can you review this too? This was added way back in 2014 and I'm pretty sure we don't use it anywhere anymore.

SamSaffron:
Yeah we can junk it, we don’t need this complexity

PR #505, opened 2020-12-18
Allow specifying tag when building image

wryun:
When I tried to specify a tag (e.g. v2.6.0) the checkout failed. --tags should fix, I believe.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -90,7 +90,7 @@ run:
         - git clean -f
         - git remote set-branches --add origin master
         - git remote set-branches origin $version
-        - git fetch --depth 1 origin $version
+        - git fetch --tags --depth 1 origin $version
         - git checkout $version
         - mkdir -p tmp
         - chown discourse:www-data tmp


SamSaffron:
@xfalcox  thoughts?

xfalcox:
> @xfalcox thoughts?

The PR makes the repository double in size during the rebuild, as it will fetch every tag leaf ever.

This puts more GitHub pressure during rebuilds making them slower, so the tradeoff is complicated.

I posted a work-around for people who want to fetch tags at https://meta.discourse.org/t/shallow-git-fetch-regression-in-discourse-docker/172324/6?u=falco

I know it sucks we lost the ability to fetch any tag out of the box, but there is a performance trade-off that 99% of the instances are using branches and don't need the tags, and are saving hundreds of megabytes due to the new way we handle the repo.

@SamSaffron  thoughts ?

SamSaffron:
Yeah I think the workaround is fine here, no need to pull this into the image. 

I guess we could have some sort of conditional support in the templates (if X is specified THEN pull the tag) to make this a bit easier. 

But ... I think the compromise we made here is correct. No need to pull all tags.

PR #512, opened 2020-12-30
Support maxmind

pfaffman:
Long-awaited discourse-setup support for maxmind keys.

If they have an old template with no default value, one gets added, though not with the helpful link to discussion that the new template has.

Oh. Darn. This also includes code to skip  check_port() if there's no netstat. I intend to replace that test that uses netstat with lsof instead, which I think it more universally included (it's installed in AWS Ubuntu which does not include netstat, anyway).

Also still to do is remove the broken disk space check since launcher does it anyway.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -381,6 +381,8 @@ ask_user_for_config() {
   if [ "$maxmind_license_key" == "1234567890123456" ]
   then 
       local maxmind_status="ENTER to continue without MAXMIND GeoLite2 geolocation database"
+  else
+      local maxmind_status="Removing a key requires editing yml by hand"
   fi
 
   read_config "DISCOURSE_HOSTNAME"
--- a/discourse-setup
+++ b/discourse-setup
@@ -279,15 +279,6 @@ check_ports() {
 ##
 check_port() {
 
-  if ! command -v netstat &> /dev/null
-  then
-    echo "netstat not found. Port check skipped."
-    echo "You should verify that no other processes are using ports 80 or 443"
-    echo "Install netstat with something like "
-    echo "    apt install net-tools"
-    return
-  fi
-  
   local valid=$(netstat -tln | awk '{print $4}' | grep ":${1}\$")
 
   if [ -n "$valid" ]; then
--- a/discourse-setup
+++ b/discourse-setup
@@ -321,7 +321,7 @@ assert_maxmind_license_key() {
   if  ! grep DISCOURSE_MAXMIND_LICENSE_KEY $web_file
   then
     echo "Adding MAXMIND placeholder to $web_file"
-    sed -i -e 's/LETSENCRYPT_ACCOUNT_EMAIL/a\ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' $web_file 
+    sed -i -e 's/LETSENCRYPT_ACCOUNT_EMAIL/a\ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' $web_file
   fi
 }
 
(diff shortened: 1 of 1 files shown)


discourse-setup
```diff
@@ -316,6 +325,17 @@ read_default() {
...
 }
 
+assert_maxmind_license_key() {
+  echo "Assert maxmind license key for $web_file"
+  if  ! grep DISCOURSE_MAXMIND_LICENSE_KEY $web_file
+  then
+    echo "Adding DISCOURSE_MAXMIND_LICENSE_KEY to $web_file"
+    sed -i -e 's/LETSENCRYPT_ACCOUNT_EMAIL/a\ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' $web_file 
+  else
+    echo "Have maxmind already"
```
riking, line 335: Since this is adding the commented-out version, this log line is misleading and should be removed. How about making the previous one "Adding .... placeholder to $web_file" ?

discourse-setup
```diff
@@ -352,6 +372,17 @@ ask_user_for_config() {
     local letsencrypt_status="Enter 'OFF' to disable."
   fi
 
+  read_config "DISCOURSE_MAXMIND_LICENSE_KEY"
+  local maxmind_license_key=$read_config_result
+  if [ -z $maxmind_license_key ]
+  then
+      maxmind_license_key="1234567890123456"
```
riking, line 379: let's use a more obvious placeholder like `INVALID_KEY`

discourse-setup
```diff
@@ -576,7 +621,7 @@ ask_user_for_config() {
     sed -i -e "s/$src/$dst/w $changelog" $web_file
     if [ -s $changelog ]
     then
-  echo "web.ssl.template.yml enabled"
+      echo "web.ssl.template.yml enabled"
```
riking, line 624: indentation problem?

discourse-setup
```diff
@@ -611,7 +670,7 @@ validate_config() {
   valid_config="y"
 
   for x in DISCOURSE_SMTP_ADDRESS DISCOURSE_SMTP_USER_NAME DISCOURSE_SMTP_PASSWORD \
-           DISCOURSE_DEVELOPER_EMAILS DISCOURSE_HOSTNAME
+           DISCOURSE_DEVELOPER_EMAILS DISCOURSE_HOSTNAME
```
riking, line 673: another odd whitespace change

riking (changes requested):

discourse-setup
pfaffman: that's a great idea, but I was using the default value that is in web_only.yml and standalone.yml. Changing it now will be complicated, as it will need to test both for 1234567890123456 and INVALID_KEY as well as changing the two existing templates.

riking: ah, didn't realize it was already there!

coding-horror:
ok @pfaffman it looks like you have feedback above, did you want to act on it? 👆

pfaffman:
Sorry I've been distracted on another project.

I should be able to get this done tomorrow.

discourse-setup
pfaffman: How about this:

discourse-setup
pfaffman: Oops. I'll remove this trailing whitespace.

discourse-setup
pfaffman: There is some other odd indentation. Some editor that I use, for example, is indenting 4 spaces after a `then` but only 2 spaces after an `else`. There are other examples of 2 vs 4 space indents. 

Perhaps I (or someone else?) should do another fix-the-formatting PR after we've worked out functional changes in this commit. I'm not clear if there are prescribed standards for formatting `bash` files.

pfaffman:
I think that this addresses the earlier comments.

Once this is accepted, I'll remove the dependency on netstat and use `lsof` instead.

discourse-setup
riking: lgtm

riking (approved):

PR #516, opened 2021-01-27
FIX: discourse-setup was not adding MAXMIND key

pfaffman:
Oops. The line that was supposed to add the MAXMIND line to app.yml was broken.

The test is to get an `app.yml` that does NOT have `DISCOURSE_MAXMIND_LICENSE_KEY` in it, run `discourse-setup` (you can ^C after the `Stopping existing container in 5 seconds or Control-C to cancel.` message--you don't need to run the whole script) and then make sure that `#DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456` actually got added.

Sorry about that, folks.

Sigh. And this is broken too.

`containers/app.yml` in the changed line should be `$web_file` and I can't figure out how to un-submit this or do another commit from the web interface where I thought I'd do this Real Quick.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -320,7 +320,7 @@ assert_maxmind_license_key() {
   if  ! grep DISCOURSE_MAXMIND_LICENSE_KEY $web_file
   then
     echo "Adding MAXMIND placeholder to $web_file"
-    sed -i '/^.*LETSENCRYPT_ACCOUNT_EMAIL.*/a \ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' containers/app.yml
+    sed -i '/^.*LETSENCRYPT_ACCOUNT_EMAIL.*/a \ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' $web_file
   fi
 }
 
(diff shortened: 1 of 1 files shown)


pfaffman:
OK. It looks like I managed to commit the correct version.

pfaffman:
OK. Hold off on this. After considering elaborate validation suites it occurred to me that all that needs to happen is have `assert_maxmind_license_key()` actually do what its name implies and include a test AFTER it's "added" the required line to see that it worked. I hope to have this done in the next couple hours.

pfaffman:
Oh. I think this is ready to go. This got buried in a commit note...

Ouch. This was worse than I thought.

Things fixed:

- consistent formatting (mostly that after `then` was indented by 4, not 2)
- fail if `DISCOURSE_MAXMIND_LICENSE_KEY` is not added to $web_file
- detect if `web_only.yml` exists 
- stop `web_only` if it exists rather than non-existing `app` (but why is it stopping it anyway?-- Because the port test fails if Discourse is running)
- don't try to `assert_maxmind_license_key` before `app.yml` exists

xfalcox (approved):

pfaffman:
Can someone get this merged? I would really like for discourse-setup not to be broken and to make further modifications that will depend on these.

PR #525, opened 2021-02-11
Add notification email to discourse setup

pfaffman:
This is primarily to add ability to set SMTP_DOMAIN and NOTIFICATION_EMAIL in `discourse-setup`. 

@xfalcox asked for this:

- Add a new question to discourse-setup: “What is the email where notifications should come from?” or something like that.
- User replies, let’s say noreply@example.com.
- You split on the @ and set example.com to SMTP_DOMAIN.
- You use the full value and set the notification_email.

This does the above, and will add the required fields to reasonably old existing config files.

My last pull request was broken if it was re-run for existing sites, so this time I have added some basic smoke tests. As part of that I also added some command line switches to `--skip-rebuild` (that works) and `--skip-connection-test` (so we can run this on a host that isn't on the internet). (I also added `--two-container` and deprecated the rather ugly `2container` option).

The tests just run `discourse-setup` and test that values got set or changed as expected. I actually caught a couple bugs!

I also changed the port check to use `nc` rather than `netstat`, which is not included in some distributions.

PR #527, opened 2021-02-22
feature:  discourse-setup email config improvements

xfalcox:
Rebased version of #525

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -2,6 +2,34 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
 
+if [ "$1" == "2container" ] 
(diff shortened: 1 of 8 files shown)


xfalcox:
@pfaffman I extracted the single commit with changes from the 28 of the #525 PR here. Can you check if those are the changes we should merge?

pfaffman:
Thanks, Rafael!

```
./tests/run-all-tests

```
Does not print failure, so it looks good!

Oh. Wait. The `replace-netcat-with-lsof` bit got lost.

Arguably that should have been another PR anyway. I think you should take this, which does what the branch name implies, and I'll do another PR for replacing `netcat` with `lsof`

davidtaylorhq (approved):

PR #528, opened 2021-02-22
FIX: discourse-setup stop running container

pfaffman:
This got commented out during testing of the previous commit. 

And remove some trailing whitespace

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -2,7 +2,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
 
-if [ "$1" == "2container" ] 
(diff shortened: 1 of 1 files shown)


SamSaffron (approved):

SamSaffron:
Sure , looks good to me, but will this `stop` explode on the first run? do we need to catch some sort of error there?

pfaffman:
On first run, we don't need to do anything. That works fine.

This test is inside a `if app-file-already-exists` condition.

If `discourse-setup` is run to reconfigure an existing instance, we need to stop the existing container, otherwise, the checks that the host name resolves to the VM fails. That's the way it's been until I commented out the `sleep` and `stop` when I was debugging.

PR #535, opened 2021-04-10
no longer allow nested templates

featheredtoast:
Drop support for nested templates. Require strict composition of templates from base app.yml files.


fitzy101:
Change looks good, though I'm curious as to why we're dropping support for this? Might be good to include that in the commit message.

fitzy101 (approved):

PR #539, opened 2021-04-28
FIX: prevent exit if themes:update errors out

featheredtoast:
Make themes:update task a best-effort - If a theme becomes unavailable to update (eg, repository deleted, moved, or pubkey deleted), we *probably* don't want to fail the build for this alone.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -165,11 +165,18 @@ run:
       hook: db_migrate
       cmd:
         - su discourse -c 'bundle exec rake db:migrate'
+
+  - exec:
+      cd: $home
+      cmd:
+        - su discourse -c 'bundle exec rake themes:update'
+      raise_on_fail: false
+
   - exec:
       cd: $home
       hook: assets_precompile
       cmd:
-        - su discourse -c 'bundle exec rake themes:update assets:precompile'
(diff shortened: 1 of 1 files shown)


SamSaffron:
The double boot of rails makes me a bit uneasy here.

Maybe we allow for `RAISE_ON_THEME_UPDATE_ERROR=false` in core? That way we can keep to 1 Rails boot? 

For context in digital ocean:

```
root@samsaffron-samsaffron:/# time rails r 'puts'

real	0m12.200s
user	0m11.192s
sys	0m0.852s

``` 

12 extra seconds for each command that boots rails in digital ocean. 

cc @OsamaSayegh / @pmusaraj

pmusaraj:
How about running `themes:update` with `plugin:pull_compatible_all` instead? That wouldn't add an additional rails boot (and it already includes the `raise_on_fail: false` flag).

SamSaffron:
Or maybe we add a "superset" task for 

`plugin:pull_compatible_all` (ignore failures) `themes:update` (ignore failures) and `db:migrate` 


`rake discourse:update` or some other name like that? 

Then we wipe out 1 rails boot which is awesome, in fact maybe we even fold precompile into this new magical task, then we wipe out 2 rails boots!

PR #543, opened 2021-06-12
FIX: `/var/lib/docker` doesn't exist on macOS

CvX:
Even though `docker info --format '{{.DockerRootDir}}'` returns that path.

### DIFF
--- a/launcher
+++ b/launcher
@@ -235,8 +235,7 @@ check_prereqs() {
   # 7. enough space for the bootstrap on docker folder
   folder=`$docker_path info --format '{{.DockerRootDir}}'`
   safe_folder=${folder:-/var/lib/docker}
-  test=$(($(stat -f --format="%a*%S" $safe_folder)/1024**3 < 5))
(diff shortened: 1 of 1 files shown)


launcher
```diff
@@ -235,8 +235,7 @@ check_prereqs() {
   # 7. enough space for the bootstrap on docker folder
   folder=`$docker_path info --format '{{.DockerRootDir}}'`
   safe_folder=${folder:-/var/lib/docker}
-  test=$(($(stat -f --format="%a*%S" $safe_folder)/1024**3 < 5))
-  if [[ $test -ne 0 ]] ; then
+  if [[ -d $safe_folder && $(gstat -f --format="%a*%S" $safe_folder)/1024**3 -lt 5 ]] ; then
```
SamSaffron, line 238: gstat is not super available, not installed on linux by default.

CvX: Ah, that wasn't supposed to make it to the commit! Good catch!

SamSaffron (approved):

PR #552, opened 2021-08-02
Update to NodeJS 16

runlevel5:
Because version 15 is no longer supported.


SamSaffron (approved):

SamSaffron:
Sure this seems like a good idea! Thanks

PR #553, opened 2021-08-02
[base image] Specify full docker.io URL

runlevel5:
It is a good practice to be explicit on which source the images are fetched from

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -1,6 +1,6 @@
 # NAME:     discourse/base
 # VERSION:  release
-FROM debian:buster-slim
+FROM docker.io/library/debian:buster-slim
 
 ENV PG_MAJOR 13
 ENV RUBY_ALLOCATOR /usr/lib/libjemalloc.so.1


SamSaffron:
@saj thoughts? should we be doing this in our images?

saj:
_Why?_  'It is good practice' is meaningless on its own.  Please help us understand why you want this patch merged.

See https://github.com/moby/moby/pull/20947

> The old examples no longer worked due to changes in the client and Docker Hub.

> Please remove the domain part (docker.io) entirely (probably library/ as well)
>
> Reviewing all docs to remove any (bogus!) domain name like registry.hub.docker.com or docker.io should really be done as well...

I don't have time to dig up the patches, but it smells like Docker have broken backwards compat on these services before.  They seem to be reliant on the implicit behaviour baked into the `docker` program.  I note that the documentation for [`docker pull`](https://docs.docker.com/engine/reference/commandline/pull/) no longer mentions any service address for Docker Hub.

runlevel5:
@saj because it could built not only with docker but other runtime such as podman. For platforms that docker are not available, for example linux/ppc64le, users have no other choice but to rely on non-docker alternatives. And most of them don't assume implicitly that docker.io is the default registry. So it is to me being explicit won't break anything with docker yet making life easier for DevOps that maintain infrastructure that does not rely on docker.

saj:
Most of our users are on Docker CE.

> So it is to me being explicit won't break anything with docker

On the contrary:  if Docker (the commercial company) have broken backwards compat with their registry services in the past, we have to assume they will do it again.  If the code and documentation that Discourse provide to our users had included a fully-qualified registry address at the time of the change mentioned in my last post, our users on Docker would, apparently, have encountered some problems.

I do feel for you.  Implicit registry addresses are a tedious special case, and I know I've burned time on this myself.  But, regrettably, the implicit registry addresses do form a part of the current 'container ecosystem' -- for lack of a better term.  **The onus is on third-party systems (including podman) to account for this existing pattern.**

https://developers.redhat.com/blog/2019/02/21/podman-and-buildah-for-docker-users

> you can pull the latest Fedora image from the Docker hub and run it using Podman.  Not specifying a registry means Podman will default to searching through registries listed in the `registries.conf` file, in the order in which they are listed. An unmodified `registries.conf` file means it will look in the Docker hub first.

It sounds like Red Hat provide sensible defaults.  Have you changed them?  Something like this should do:

```
[registries.search]
registries = ['docker.io']
```

See [`containers-registries.conf(5)`](https://github.com/containers/image/blob/2fefc38386d53f639186cbd6c10a0544864fe1b2/docs/containers-registries.conf.5.md).

xfalcox:
> For platforms that docker are not available, for example linux/ppc64le

Worth mentioning that Discourse only works in x86_64 at the moment. There is work in a upstream library for it to work in arm64 someday, but I doubt it will work in that powerpc.

runlevel5:
> It sounds like Red Hat provide sensible defaults. Have you changed them? Something like this should do:

Yes I am aware of the `unqualified-search-registries` config.

The thing is that the sensible default varies by Linux distro. For example latest Fedora does have docker.io as fallback registry but there is no guarantee some distro would be the same.

> Worth mentioning that Discourse only works in x86_64 at the moment. There is work in a upstream library for it to work in arm64 someday, but I doubt it will work in that powerpc.

It is already working with ppc64le :)

runlevel5:
I do agree with you the adding the full registry URL might be introducing too much changes across the documentation and source to optimise for edge case of non-docker runtime. Perhaps sticking to the de-factor docker is still the right choice as it is much easier to support. Great thanks for sharing your insight with me. Keep up the good work.

PR #558, opened 2021-08-12
Fix line break handling in Cloudflare template

Zenexer:
Cloudflare's IP list has gone back and forth between including a trailing line break and omitting it.

When a trailing line break was first added in 2015, it resulted in a bug: https://meta.discourse.org/t/issue-with-cloudflare-template/35113

The trailing line break was removed again in 2021: https://meta.discourse.org/t/cloudflare-template-broken-again/200219

This fixes the template so that it will work regardless of extra line breaks.  It will also safely ignore any empty lines that may appear in the files.

### DIFF
--- a/templates/cloudflare.template.yml
+++ b/templates/cloudflare.template.yml
@@ -6,9 +6,10 @@ run:
         #!/bin/bash -e
         # Download list of CloudFlare ips
         wget https://www.cloudflare.com/ips-v4/ -O - > /tmp/cloudflare-ips
+        echo >> /tmp/cloudflare-ips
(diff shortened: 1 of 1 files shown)


FaizanZahid:
We are having same issue, it this related to same bug?
https://meta.discourse.org/t/discourse-updated-from-admin-not-working-after-rebuild-fatal-error-site-does-not-load-anymore-after-rebuild/200443

chiefy:
Just ran into this as well, thanks for the fix!

xfalcox (approved):

PR #572, opened 2021-10-04
DEV: replace mailcatcher with mailhog

techAPJ:
Accompanying PR in core: https://github.com/discourse/discourse/pull/14549

### DIFF
--- a/image/discourse_dev/Dockerfile
+++ b/image/discourse_dev/Dockerfile
@@ -5,10 +5,6 @@ ENV RAILS_ENV development
 
 #LABEL maintainer="Sam Saffron \"https://twitter.com/samsaffron\""
 
-# Install for mailcatcher gem
-RUN apt update && apt install -y libsqlite3-dev \
-    && gem install mailcatcher && rm -rf /var/lib/apt/lists/*
-
 # Remove the code added on base image
 RUN rm -rf /var/www/*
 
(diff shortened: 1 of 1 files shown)


image/discourse_dev/Dockerfile
```diff
@@ -5,9 +5,9 @@ ENV RAILS_ENV development
 
 #LABEL maintainer="Sam Saffron \"https://twitter.com/samsaffron\""
 
-# Install for mailcatcher gem
-RUN apt update && apt install -y libsqlite3-dev \
-    && gem install mailcatcher && rm -rf /var/lib/apt/lists/*
+# Install for MailHog (https://github.com/mailhog/MailHog)
+RUN apt update && apt install -y golang-go \
+    && go get github.com/mailhog/MailHog
```
davidtaylorhq, line 10: This doesn't make mailhog available on the PATH. To run it, people would have to type

```
/root/go/bin/MailHog ...
```
(plus I'm not sure if there would be permissions problems, since the binary is build as `root`, and might be run as `discourse`)

At a minimum we should move the binary to `/usr/local/bin`. Even better would be to copy the 'multistage build' technique from [MailHog's own Dockerfile](https://github.com/mailhog/MailHog/blob/master/Dockerfile). The added bonus there is that we don't have to ship a golang development environment in our docker image.

SamSaffron: I would not install the full go toolset, just copy the binary to the right place

davidtaylorhq: At the moment the golang toolset is required to build the binary in the first place. If we're happy to hard-code the MailHog version, then we could fetch the pre-built `linux_amd64` binary from https://github.com/mailhog/MailHog/releases/

techAPJ: I've made the required change.

image/discourse_dev/Dockerfile
```diff
@@ -44,3 +40,7 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo ap
     apt update &&\
     apt install -y google-chrome-stable firefox-esr &&\
     npm install -g eslint babel-eslint
+
+# Install & Configure MailHog (https://github.com/mailhog/MailHog)
+RUN wget -nc -qO /usr/local/bin/mailhog https://github.com/mailhog/MailHog/releases/download/v1.0.1/MailHog_linux_amd64\
```
SamSaffron, line 45: We should be in the habit of doing a SHA check when we download stuff, just in case.

PR #577, opened 2021-11-04
added github.china template to use github mirror to bypass firewall in China

hamo:
In China, Github is sometimes blocked by firewall, added a new template to use GitHub mirror hub.fastgit.org to bypass it.


HuengchI:
There's already a `hooks before_code` statement in the `redis.template.yml`.
I don't know if multiple `hooks` will override each other.

PR #583, opened 2021-11-29
support:real IP in ssl proxy 

qyh214:
https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol


SamSaffron:
Not sure about this, it copies and pastes too much from other templates. Recommend you simply do a meta topic about this instead.

PR #584, opened 2021-12-02
Build aarch64 image

xfalcox:


image/base/Dockerfile
```diff
@@ -10,6 +10,7 @@ ENV RAILS_ENV production
 
 RUN echo 2.0.`date +%Y%m%d` > /VERSION
```
CvX, line 11: Why it doesn't like that line? 🤔 

```
#6 [ 2/54] RUN echo 2.0.`date +%Y%m%d` > /VERSION
#6 0.278 standard_init_linux.go:228: exec user process caused: exec format error
#6 ERROR: executor failed running [/bin/sh -c echo 2.0.`date +%Y%m%d` > /VERSION]: exit code: 1

error: failed to solve: executor failed running [/bin/sh -c echo 2.0.`date +%Y%m%d` > /VERSION]: exit code: 1
```

xfalcox: Oh that is expected :sweat_smile: 

Gotta setup the whole emulator before hand, still WIP

CvX:
2h 12m and still going strong! 😁

PR #588, opened 2021-12-13
Introduce a discourse/base:slim image

davidtaylorhq:
The base-slim image doesn't run `yarn install` or `bundle install`, so it is much more lightweight. This is intended for use by systems which do their own dependency caching (e.g. GitHub Actions)

The image will be released under the `:slim` tag, and also `:v2.0.{timestamp}-slim`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,6 @@ on:
 
 jobs:
   base:
-    strategy:
-      matrix:
-        name: ["base", "base_slim"]
     runs-on: [ubuntu-20.04]
     steps:
       - name: enable experimental docker features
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -20,24 +23,36 @@ jobs:
...
       - name: build base image
         run: |
-          cd image && ruby auto_build.rb base
+          cd image && ruby auto_build.rb ${{ matrix.name }}
       - name: build test_build image
+        if: matrix.name == 'base'
         run: |
           cd image && ruby auto_build.rb discourse_test_build
       - name: run specs
+        if: matrix.name == 'base'
```
CvX, line 32: Does this mean that in the case of failing specs, it won't publish `base:release` but **will** go ahead with `base:slim`? 🤔

davidtaylorhq: Yes that is true. Let me see if I can make it install the dependencies and run the specs 👀

image/auto_build.rb
```diff
@@ -4,7 +4,8 @@
 require 'optparse'
 
 images = {
-  base: { name: 'base', tag: "discourse/base:build", squash: true },
+  base_slim: { name: 'base', tag: "discourse/base:build_slim", squash: true, extra_args: '-f slim.Dockerfile' },
```
xfalcox, line 7: Where does this Dockerfile lives?

xfalcox (changes requested):

image/auto_build.rb
davidtaylorhq: 🤦 Now pushed both Dockerfiles

PR #592, opened 2021-12-14
DEV: Fix aarch64 build

davidtaylorhq:
buildx can't build `FROM` a local image, so we need to set up a temporary local registry for the intermediate image


xfalcox (approved):
So many extra lines :sob:

davidtaylorhq:
Yeah I know 😢 

The ideal solution is to make the base image a multi-stage build (like I did with discourse_test). Unfortunately if we do that, we can't use `--squash`, which is pretty important for the base image at the moment. A lot of refactoring required before we can get to that point.

PR #597, opened 2022-01-25
UX: launcher time out if no reply on cleanup

pfaffman:
When doing a `read` to see if they want to remove images or postgres backups, time out if no response in 60 seconds.

If you're calling `./launcher` from a script (e.g., Ansible) and the disk is full for some reason other than Docker or postgres backups, `./launcher` will hang forever.


tgxworld:
Closing as this PR has gone stale. Please feel free to open a new PR and ping me if it is still relevant.

PR #603, opened 2022-02-11
DEV: Update phpBB3 template to work with latest base image

gschlager:


templates/import/phpbb3.template.yml
```diff
@@ -36,7 +36,7 @@ hooks:
     - exec:
         cmd:
           - mkdir -p /shared/import/mysql/data
-          - apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y nano libmariadbclient-dev mariadb-server
+          - apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y nano libmariadb-dev mariadb-server
```
CvX, line 39: What's the story behind this change? Do we need a full db, or did they rename/merge packages, or something else? 😃

CvX (approved):

pfaffman:
I noticed on a site I was working on this week that `vim` was also missing, maybe add it in addition to `nano`?

Or maybe people who know vim know how to install it. :-)

templates/import/phpbb3.template.yml
gschlager:
```
Package libmariadbclient-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libmariadb-dev-compat libmariadb-dev
```

gschlager:
> Or maybe people who know vim know how to install it. :-)

I'm sure they know otherwise they should use `nano` :wink:

PR #610, opened 2022-03-10
DEV: use consistent third-party installation method

gabenp:
* Add checksum verification where it was missing for external deps
  * jemalloc
  * oxipng
* Remove unnecessary package installations, move some package installations around
  * notably `software-properties-common` was removed, it was adding unnecessary packages
* `wget` output is quieted as to not spam CI logs
* Replace some file ADDs with COPY using nested directories in the repo (see `etc/` and `sbin/`)


image/base/slim.Dockerfile
```diff
@@ -59,47 +58,30 @@ RUN cd / &&\
...
-    rm -rf /usr/local/share/ri/2.7.5/system
-
-ADD install-redis /tmp/install-redis
-RUN /tmp/install-redis
-
-ADD install-imagemagick /tmp/install-imagemagick
-RUN /tmp/install-imagemagick
-
-# Validate install
-RUN ruby -Eutf-8 -e "v = \`convert -version\`; %w{png tiff jpeg freetype heic}.each { |f| unless v.include?(f); STDERR.puts('no ' + f +  ' support in imagemagick'); exit(-1); end }"
```
CvX, line 102: Just to point out: this PR keeps ImageMagick install validation, but no longer has an explicit step of validating ruby installation. That line kinda did both, but I think we're fine.

CvX (approved):

image/base/slim.Dockerfile
gabenp: Ah I didn't consider that when I replace it. The `gem` steps immediately below the ruby install should also do a rudimentary validation of the ruby install.

I just dug a bit further on my CI runs here and noticed in the base `test_build` step:
```
 gem install net-http-persistent -v 4.0.1', require_name: 'net/http/persistent -i /var/www/discourse/plugins/discourse-algolia/gems/2.7.5 --no-document --ignore-dependencies --no-user-install
ERROR:  While executing gem ... (Gem::Requirement::BadRequirementError)
    Illformed requirement ["require_name: net/http/persistent"]
```

It's been like that for months as seen here https://github.com/discourse/discourse-algolia/blame/c57118ba425fecf9da33d77fb98e511b5fd88f98/plugin.rb#L10

It doesn't seem to be breaking the processes here for the base image (specs etc) but it's a bit odd and seems worth mentioning. I'm not sure exactly what it's trying to accomplish there.

CvX: Fixed in https://github.com/discourse/discourse/pull/16192 😃

PR #611, opened 2022-03-10
FIX: pg_basebackup parameter in take-database-backup

Firefishy:
`take-database-backup` script tries to use `pg_baseback --xlog` parameter which was remove in postgres >=10.

https://www.postgresql.org/docs/10/app-pgbasebackup.html


Firefishy:
@CvX @gabenp Any changes needed? The current script does not function as the `--xlog` doesn't exist in recent versions of postgres.

gabenp:
This looks fine to me. I approved the CI builds and given no issues seen there (don't expect any) then we can merge.

Thanks for the contribution!

gabenp (approved):

PR #615, opened 2022-03-31
add container build time

Supermathie:
- Cleanup code in the bootstrap function
- Add label showing image creation time on commit

### DIFF
--- a/launcher
+++ b/launcher
@@ -140,6 +140,10 @@ compare_version() {
     return 1
 }
 
+die () {
+  echo -e "\n$1\n" >&2
+  exit "${2:-1}"
+}
 
 install_docker() {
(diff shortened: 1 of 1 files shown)


Supermathie (commented):
maybe better named as `fatal`

everywhere else we spit our error messages to STDOUT, we should probably do the same here instead of STDERR

gabenp (approved):

andrewschleifer (approved):

andrewschleifer:
> everywhere else we spit our error messages to STDOUT, we should probably do the same here instead of STDERR

The eternal dilemma: Be inconsistent and correct or consistent and _wrong_.

(Unless it's egregious, I generally go with consistency, because making ALL THE THINGS correct later is often easier than finding which subset needs fixing.)

PR #617, opened 2022-04-08
DEV: add pg and redis gem for protocol-aware healthcheck

fitzy101:
The cache_critical_dns script is going to be updated to perform a
healthcheck on the Redis and Postgres hosts prior to caching the
hostname and IP address resolved via DNS. In order to perform a
protocol-aware healthcheck (i.e. a PG/Redis PING-PONG transaction) the
pg and redis gems are required. Ensure the gems are only installed if
the DNS caching behaviour is enabled via templates.

### DIFF
--- a/templates/cache-dns.template.yml
+++ b/templates/cache-dns.template.yml
@@ -15,5 +15,6 @@
 
 hooks:
   after_bundle_exec:
+    - exec: "gem install pg redis"
     - exec: "mkdir -p /etc/service/cache_critical_dns/"
     - exec: "cp /var/www/discourse/script/cache_critical_dns /etc/service/cache_critical_dns/run"


templates/cache-dns.template.yml
```diff
@@ -15,5 +15,6 @@
 
 hooks:
   after_bundle_exec:
+    - exec: "gem install pg redis"
```
xfalcox, line 18: Ain't those gems already installed because they are listed in the Discourse Gemfile? And it's not like we will move away from pg/redis.

fitzy101: I thought that would be the case, but they aren't available as the user that runs this script (`root`). They also aren't listed via `gem list`. Probably installed as a different user? 

In any case, best scenario is this command is a no-op, worst case is that gems are installed.

CvX: Discourse app gems are bundled locally, see:

https://github.com/discourse/discourse_docker/blob/2a9faf7e5680b9979585be613162c45a0960a525/image/base/release.Dockerfile#L7-L8

So this will install those two again, in a different place

andrewschleifer: Even though it's shipped with Discourse core, the `cache_critical_dns` script won't use any of the vendored libraries. It uses the ["system" ruby](https://github.com/discourse/discourse/blob/650adbe423ca2ed2394dd06054d50052a91e9d08/script/cache_critical_dns#L1) (and gems) in the container.

fitzy101: @CvX @xfalcox would you mind either approving or suggesting a change, given the explanation for installing the gems?

PR #627, opened 2022-05-02
DEV: Ensure secure file permissions by default

gabenp:
`discourse-setup` will now ensure container definitions are
installed with `0600` permissions mode only.

`launcher` will now throw a warning when an existing container
definition is world-readable.

Also clean up leftover `launcher setup` logic which no longer exists.
Merge pre-existing logic into `check_prereqs` function.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -884,7 +884,7 @@ then
   DATE=`date +"%Y-%m-%d-%H%M%S"`
   BACKUP=$app_name.yml.$DATE.bak
   echo Saving old file as $BACKUP
-  cp $web_file containers/$BACKUP
+  install -m0600 $web_file containers/$BACKUP
   if [ "$DEBUG" != "1" ]
   then
     echo "Stopping existing container in 5 seconds or Control-C to cancel."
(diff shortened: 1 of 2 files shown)


andrewschleifer (approved):

pfaffman:
If it's important for containers to be `0600` all of a sudden, maybe it would be OK to just set them?

Most people who run `./launcher` don't know what permissions are.

If it's not appropriate for `launcher` to change the permissions, would it be OK to at least have `discourse-setup` set them for people? Adding one more mysterious command to the install sequence seems unnecessary.

CvX:
> would it be OK to at least have discourse-setup set them for people?

Doesn't the change in `discourse-setup` handle this?

CvX:
@gabenp Please take a look at that topic ^ - can we automatically handle permissions issues like that one? 😃

pfaffman:
A thousand apologies! If I'm to offer criticism I should at least pay enough attention that I see what's going on.

I think that all makes sense. My concern started when someone noticed that warning and because of a whole bunch of other stuff that looks sort-of like warnings he went back to try to understand the rest of the stuff, at least some of which, I think, is spurious. I'm pretty sure that discourse-setup prints out a bunch of stuff that's part of a test, which is confusing (and I wrote it, so I am not blaming anyone else).

But I think I should just hush on this one.

Thanks very much.

PR #630, opened 2022-05-05
Create sqlite-dep.template.yml

pfaffman:
Add template for those times when you just need sqlite3.

Doh! Should I have named it sqlite3-dep.template.yml? (EDIT: Well, I googled a bit and it seems sqlite3 is just the current version of sqlite, so I think the suggested filename is appropriate. :grin: )


SamSaffron:
I am just not sure about this, the more little snowflakes we keep the more things we need to maintain.

PR #643, opened 2022-09-25
Ubuntu Jammy (22.04) doesn't have libwebp6

skynavga:
Conditionally use libwebp7 on Jammy. N.B. This conditionalization should probably be generalized.


xfalcox (approved):

xfalcox:
Thanks for the PR @skynavga !

PR #648, opened 2022-12-01
Update Ruby to 3.1

xfalcox:


CvX:
imagemagick/oxipng bump breaks a spec

CvX:
Actually, that's `pngquant` which we don't pin 😅 I'm updating the spec in core.

CvX (approved):

PR #649, opened 2022-12-20
fix history

xfalcox:
- Revert "bump deps"
- Revert "Update Ruby to 3.2"
- Reland Ruby 3.1.3 and deps bump


tgxworld:
Minor suggestion, instead of "Reland Ruby 3.1.3 and deps bump", can we just say "Update Ruby to 3.1.3 and deps bump"? I don't think there is a need to hold reference to the fact that we're doing this again.

xfalcox:
> Minor suggestion, instead of "Reland Ruby 3.1.3 and deps bump", can we just say "Update Ruby to 3.1.3 and deps bump"? I don't think there is a need to hold reference to the fact that we're doing this again.

Done!

tgxworld (approved):

tgxworld:
@xfalcox Looks good :+1:

PR #659, opened 2022-12-30
Install correct bundler version for dev

xfalcox:


gschlager:
Isn't bundler supposed to install the correct version of bundler automatically? I've seen it install the version from `Gemfile.lock` multiple times recently when a different version of bundler was running. :thinking:

davidtaylorhq:
Yeah, modern version of bundler will auto-install the version from the lockfile. I wonder if the permissions errors we're seeing are because our original install of bundler is done as root, and then in the dev image we switch to the `discourse` user?

xfalcox:
Yes, that is exactly the case which is why I'm pre-installing the correct version of bundler in a step before as root.

davidtaylorhq:
But then if someone's using the dev image, and they pull the latest discourse core without updating the image, they would run into the same issue?

davidtaylorhq:
This should do it: https://github.com/discourse/discourse_docker/pull/662

PR #676, opened 2023-02-09
bash tab completion for launcher + discourse-setup

pfaffman:
add tab completion for launcher and discourse-setup.

For launcher, offers command (e.g., rebuild, start) and then offers yml files from containers directory. After that switches (e.g., --run-image) are offered. (Will not offer switches except in final position, sorry.)

discourse-setup offers switches (e.g., --two-container).

discourse-docter has no command line arguments.


andrewschleifer (approved):

pfaffman:
I don't think this works. It's hard to debug since you need to start a new bash each time to test, I've subsequently learned. It does work to tab-complete commands, but not the  `yml` files. Since this has gotten attention, I'll take another quick stab at it. It would be very handy.

PR #680, opened 2023-02-14
Fix: Remove egrep for grep -E

fe80:
Hello,

Egrep is deprecated since [08/2011](https://git.savannah.gnu.org/cgit/grep.git/commit/src/egrep.sh?id=a9515624709865d480e3142fd959bccd1c9372d1). You need to change with `grep -E`

Regards,


tgxworld:
Thank you for your contribution here @fe80 and apologies for the very late review.

tgxworld (approved):

PR #683, opened 2023-02-26
Fix issue with "The source cluster was not shut down cleanly"

Corillian:
This resolves the problem with database upgrades failing with the error message `The source cluster was not shut down cleanly`. The resolution comes from https://github.com/tianon/docker-postgres-upgrade/issues/27.


tgxworld:
Apologies for the late review but it looks like this PR has gone stale. Please feel free to create a new PR and ping me if you'll still like to get it reviewed.

PR #684, opened 2023-02-28
FEATURE: Install pgvector for PostgreSQL

xfalcox:


image/base/install-pgvector
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+OXIPNG_VERSION="0.4.0"
```
oblakeerickson, line 4: should this be here?

oblakeerickson (changes requested):

oblakeerickson (approved):

PR #688, opened 2023-03-08
Compile oxipng from source

djlambert:
Download oxipng source from GitHub and compile it.

Also uses v8.0.0 instead of v7.0.0. There is a breaking change in the changelog, "Revamp alpha optimization", presumably that only impacts the files generated.

I'm not sure if it's still necessary to install the deps as Cargo pulls down what it needs, but leaving.


SamSaffron (approved):

SamSaffron:
thanks heaps!

PR #689, opened 2023-03-09
DEV: Eliminate warnings and noise during image build

djlambert:
This eliminates most of the warnings in the build process from `apt` and `apt-key`. It also disables the progress output from a couple `wget` and `curl` commands which weren't already.

Removed the deprecated option `--with-ipv6` for nginx.

### DIFF
--- a/image/base/install-imagemagick
+++ b/image/base/install-imagemagick
@@ -15,15 +15,15 @@ PREFIX=/usr/local
 WDIR=/tmp/imagemagick
 
 # Install build deps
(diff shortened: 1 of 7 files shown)


djlambert:
Not sure why the aarch64 build is failing. I don't see the same issue on my M1 Mac.

tgxworld:
Apologies for the late review but it looks like this PR has gone stale. Please feel free to create a new PR and ping me if you'll still like to get it reviewed.

PR #691, opened 2023-03-09
DEV: Create partial clone of discourse repository

gschlager:
A treeless clone is nearly as small as a shallow clone, but it contains the full history which makes it possible to easily checkout tags and commits and allows all features of docker_manager to work correctly.

See https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ for more information on treeless clones.

Also this change makes it possible to correctly pull the main branch which didn't work before.

### DIFF
--- a/image/base/slim.Dockerfile
+++ b/image/base/slim.Dockerfile
@@ -122,5 +122,4 @@ COPY sbin/ /sbin
 # Discourse specific bits
 RUN useradd discourse -s /bin/bash -m -U &&\
     install -dm 0755 -o discourse -g discourse /var/www/discourse &&\
-    sudo -u discourse git clone --depth 1 https://github.com/discourse/discourse.git /var/www/discourse &&\
-    sudo -u discourse git -C /var/www/discourse remote set-branches --add origin tests-passed
+    sudo -u discourse git clone --filter=tree:0 https://github.com/discourse/discourse.git /var/www/discourse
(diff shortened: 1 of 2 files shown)


image/base/slim.Dockerfile
```diff
@@ -122,5 +122,4 @@ COPY sbin/ /sbin
 # Discourse specific bits
 RUN useradd discourse -s /bin/bash -m -U &&\
     install -dm 0755 -o discourse -g discourse /var/www/discourse &&\
-    sudo -u discourse git clone --depth 1 https://github.com/discourse/discourse.git /var/www/discourse &&\
-    sudo -u discourse git -C /var/www/discourse remote set-branches --add origin tests-passed
+    sudo -u discourse git clone --filter=tree:0 https://github.com/discourse/discourse.git /var/www/discourse
```
CvX, line 125:
```suggestion
    sudo -u discourse git clone --filter=tree:0 https://github.com/discourse/discourse /var/www/discourse
```

I believe `.git` suffix is just a back-compat redirect, no need to keep that 😃

CvX (approved):

ZogStriP (approved):

image/base/slim.Dockerfile
gschlager: GitHub shows that as clone URL, so I assume that's the recommended URL. I think the redirect happens only for regular browsers, not for the git client.

PR #692, opened 2023-03-09
FIX: Remove duplicate anacron cron jobs

djlambert:
Both `cron.d_anacron` and `anacron` exist in `/etc/cron.d` with the same contents, so it's getting run twice.


tgxworld:
@djlambert Apologies for the really late review here. The fix looks good to me 👍 Will you be able to resolve the conflicts? Otherwise, I am happy to do it too.

andrewschleifer:
Yes we still need to fix that TODO.

PR #694, opened 2023-03-10
FIX: Compiling oxipng didn't work on aarch64

gschlager:
This uses the new sparse-registry feature from Rust 1.68 which lowers the memory usage. This avoids that cargo is killed due to an OOM issue (exit code 137) during "Updating crates.io index".

### DIFF
--- a/image/base/install-oxipng
+++ b/image/base/install-oxipng
@@ -18,7 +18,7 @@ echo "${OXIPNG_HASH} ${OXIPNG_ARCHIVE}" | sha256sum -c
 tar -zxf ${OXIPNG_ARCHIVE}
 cd ${OXIPNG_DIR}
 
-/usr/local/cargo/bin/cargo build --release
+CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse /usr/local/cargo/bin/cargo build --release
 cp target/release/oxipng /usr/local/bin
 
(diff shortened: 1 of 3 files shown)


image/base/install-oxipng
```diff
@@ -18,7 +18,11 @@ echo "${OXIPNG_HASH} ${OXIPNG_ARCHIVE}" | sha256sum -c
 tar -zxf ${OXIPNG_ARCHIVE}
 cd ${OXIPNG_DIR}
 
-/usr/local/cargo/bin/cargo build --release
+if [ $(dpkg --print-architecture) == "amd64" ]; then
+  /usr/local/cargo/bin/cargo build --release
+else
+  /usr/local/cargo/bin/cargo build --release --jobs 1 --verbose
+fi
```
djlambert, line 25: I found a post suggesting setting `RUST_LOG=cargo=debug`, so maybe some output from

```
RUST_LOG=cargo=debug /usr/local/cargo/bin/cargo build --release --jobs 1 --verbose
```

gschlager: Thanks! I think I fixed the issue by using a feature from the brand-new Rust 1.68. I've updated the PR and we will know for sure in a couple of hours.

djlambert (commented):
LGTM. I'll rebase my other PRs once this is merged.

CvX (approved):

PR #697, opened 2023-03-21
Removing the -p from the "nc" command.

mitkolazov:
Removing the -p from the "nc" command.
Reason:
# nc -w 4 -l -p 80
nc: cannot use -p and -l

Without -p it works just fine.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -60,7 +60,7 @@ connect_to_port () {
       esac
     done
   else
-    echo -e "HTTP/1.1 200 OK\n\n $VERIFY" | nc -w 4 -l -p $PORT >/dev/null 2>&1 &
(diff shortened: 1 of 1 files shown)


The-Doggy:
Just to add on to this, in newer versions of nc using -p with -l will actually error and cause discourse-setup to falsely report that a connection cannot be established on the given port. Whereas older versions of nc just silently ignore it and will continue on anyway.
See: [https://linux.die.net/man/1/nc](url)

> -l' Used to specify that nc should listen for an incoming connection rather than initiate a connection to a remote host. It is an error to use this option in conjunction with the -p, -s, or -z options. Additionally, any timeouts specified with the -w option are ignored.

tgxworld (approved):

PR #701, opened 2023-04-04
Add template to enable an IPv6 listener

Supermathie:
By default nginx in the container only listens for IPv4 connections. Some users
want to enable direct IPv6 connectivity to the container either via forwarding
or NAT. This provides an easy method to allow that.

### DIFF
--- a/samples/standalone.yml
+++ b/samples/standalone.yml
@@ -11,8 +11,10 @@ templates:
   - "templates/postgres.template.yml"
   - "templates/redis.template.yml"
   - "templates/web.template.yml"
+  ## Uncomment the next line to enable the IPv6 listener
+  #- "templates/web.ipv6.template.yml"
   - "templates/web.ratelimited.template.yml"
-## Uncomment these two lines if you wish to add Lets Encrypt (https)
+  ## Uncomment these two lines if you wish to add Lets Encrypt (https)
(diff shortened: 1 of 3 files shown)


davidtaylorhq (approved):

GuidoDr:
thanks a lot, I will enable it as soon as it is available and will remove my current changes.
I do really appreciate that new PR.

Firefishy:
Does this also enable IPv6 for :443?

Supermathie:
> Does this also enable IPv6 for :443?

this functionality is already in the SSL template, yes.

GuidoDr:
thanks you guys, all working fine now. 👍

Firefishy:
Splitting the `listen [::]:80;` into a separate template doesn't feel right to me when port 443 already has `listen [::]:443;` without the `web.ipv6.template.yml`

Ideally BOTH `listen [::]` should move to the ipv6 template or the 80 ipv6 listen should just be added to the base.

Firefishy:
> By default nginx in the container only listens for IPv4 connections. Some users want to enable direct IPv6 connectivity to the container either via forwarding or NAT. This provides an easy method to allow that.

This statement is not true. IPv6 was already enabled for port 443, but IPv6 was not enabled only for port 80. The new `web.ipv6.template.yml` just complicates it with half and half even further.

PR #703, opened 2023-04-10
Bump default base image to run Ruby 3.2.2 as default

tgxworld:
Ruby 3.2.2 was released with a two security fixes. While the security
fixes have been assessed to have low impact on Discourse, we still want
to get the upgrade in as a preventive measure.


OsamaSayegh (approved):

tgxworld:
Thank you @OsamaSayegh

PR #704, opened 2023-04-14
DEV: Remove HSTS duplication. Add a ssl dhparam

Firefishy:
* Remove duplication of HSTS headers. Normalise max-age to 2 years
* On-demand create a SSL dhparam file and enable ssl_dhparam
* Increase ssl_session_cache size


Firefishy:
The session cache 1MB can only store about 4000 sessions (24h). The increase to 10MB will allow around 40,000 cached sessions.

Firefishy:
Any love?

Firefishy:
Prefer if I split this up?

Firefishy:
I'll split this up. Radio silence.

Firefishy:
:-(

PR #708, opened 2023-04-18
DEV: Install all yarn dependencies

davidtaylorhq:
`devDependencies` is often used for things like ember-cli build tooling. Until now we have been moving those things to `dependencies` in our `package.json`, but it makes more sense for us to conform to the industry norms.


CvX (approved):

davidtaylorhq:
The original context for using `--production` can be found in https://github.com/discourse/discourse_docker/commit/53ef82b53910439f1af0eaef37ffb5371e0e4268

Keeping this as a draft while I ensure that the lefthook-related issue doesn't come back to haunt us 👀

CvX (approved):

PR #718, opened 2023-06-09
PERF: Uninstall rust toolkit once ruby is built

davidtaylorhq:
This tooling adds significant size to our docker image. We only need rust tooling while installing ruby, and can clean it up immediately afterwards to avoid it persisting in the image


davidtaylorhq:
Before/after: SCR-20230609-lvhnSCR-20230609-lvex

So a 580MB saving on the image size

xfalcox (approved):

PR #725, opened 2023-07-26
Move `RAILS_ENV` to release image; build discourse_dev from slim

davidtaylorhq:
Unfortunately there is no way to 'unset' an ENV in a Dockerfile. We were working around this by setting `RAILS_ENV` to an empty string in the discourse_dev dockerfile, but that didn't work in every situation.

The discourse_dev image doesn't rely on anything in the 'release' layer of the base image. In fact, it deletes the entire contents of `/var/www/*`.

This commit resolves the situation by:

- Moving the `RAILS_ENV=production` line to the 'release' layer of the base image
- Updating the discourse_dev image to be based on the 'slim' base image

This should also make the discourse_dev image much smaller, with no loss of functionality.

(ref https://github.com/discourse/discourse/pull/22765)

### DIFF
--- a/image/base/release.Dockerfile
+++ b/image/base/release.Dockerfile
@@ -3,6 +3,8 @@ ARG tag=build_slim
 
 FROM $from:$tag
 
+ENV RAILS_ENV=production
+
 RUN cd /var/www/discourse &&\
     sudo -u discourse bundle config --local deployment true &&\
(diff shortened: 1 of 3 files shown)


CvX (approved):

hawm:
Seems the `LEFTHOOK` ENV should apply the same workaround, else any git hook manage by lefthook wouldn't be trigger actually.

PR #728, opened 2023-08-24
FEATURE: add ripgrep to the base image

SamSaffron:
ripgrep is going to be used by discourse-ai to extract context from files
on the local disk.

This does increase the size of the image by 1.7MB or so, but heavily simplifies
deployment

Alternative mechanisms exists such as packing the binary in a gem but we
would pay more at install time delays.


SamSaffron:
@xfalcox / @martin-brennan thoughts?

martin-brennan (approved):

PR #732, opened 2023-09-06
FIX: Allow other commands to executed for the `discourse_test` image

tgxworld:
Why this change?

This sets the default command that is run to `ruby script/docker_test.rb` to allow
other commands to be run for the image since the `ENTRYPOINT` is no
longer `sudo -E -u discourse -H ruby script/docker_test.rb` but instead
`sudo -E -u discourse -H`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+# Use an existing docker image as a base
+FROM ubuntu
+
+# Update and install sudo
+RUN apt-get update && apt-get install -y sudo
+
+# Create a new user
+RUN useradd -m discourse
+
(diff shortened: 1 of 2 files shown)


nattsw (approved):
TY!

nattsw:
Btw, was there a topic that triggered this?

tgxworld:
@nattsw I'm working towards allowing the `discourse_test` image to be able to used in the development environment as part of our efforts to support running system tests for themes so this change is necessary such that other commands can be passed to the image.

tgxworld:
Closing this as I might not need this change.

PR #733, opened 2023-09-06
Set default user to `discourse` user

tgxworld:
There is almost no reason to set root as the default user when almost all commands have to be executed as the `discourse` user.


martin-brennan (approved):
Nice, good point!

tgxworld:
Sorry this one is a bad idea... broke a bunch of things.

PR #735, opened 2023-09-13
DEV: Install bundler version used by Discourse as a default.

tgxworld:
Why this change?

Previously we were running `gem install bundler` to install bundler but
this would install the latest bundler version instead of the bundler
version being used by Discourse. As a result, we end up having to
install bundler over and over again when using the image most of the
time.


davidtaylorhq (approved):
Change looks good to me :+1:

IMO would be nice to separate out the unrelated parts of the diff (and land them in a separate PR if needed)

tgxworld:
> IMO would be nice to separate out the unrelated parts of the diff (and land them in a separate PR if needed)

Hmm my editor has some sort of autoformat on save for `Dockerfile`s. Have removed the formatting changes from the PR.

PR #736, opened 2023-09-15
DEV: Configure bundle jobs to 4

tgxworld:
This ensures that we be default runs bundle install with at least 4
jobs.


Drenmi (approved):
4️⃣

tgxworld:
Thank you @Drenmi

PR #737, opened 2023-09-22
use DISCOURSE_BASE_URL in mail-receiver sample

evgeni:
while `DISCOURSE_MAIL_ENDPOINT` is still accepted by the mail-receiver code, the documentation prefers `DISCOURSE_BASE_URL` and so should this example

see https://github.com/discourse/mail-receiver/blob/deae52039f4679b31e01302dadace3c0f018cec8/README.md

### DIFF
--- a/samples/mail-receiver.yml
+++ b/samples/mail-receiver.yml
@@ -27,11 +27,11 @@ env:
 #  POSTCONF_smtpd_tls_security_level: may
 
 
-  ## The URL of the mail processing endpoint of your Discourse forum.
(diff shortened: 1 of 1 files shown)


pfaffman:
IF this does get changed then samples/mail-receiver.yml should also be updated.

pfaffman:
Ah. Cool. Sorry about that. I thought that this had been merged and changed my code to depend on it and then realized that my update breaks because . . . well, this hasn't been merged.

PR #738, opened 2023-10-04
Bump default base image to discourse/base:2.0.20231004-0028

andrewschleifer:


nattsw (approved):
Looks ok but where can I find where this value is? `20231004-0028`. 

EDIT: Oh https://hub.docker.com/r/discourse/base/tags ok nvm

PR #702, opened 2023-04-08
fixup count of physical cores in discourse-setup

BigEd:
Previously parsing /proc/cpuinfo but that fails on Raspberry Pi and presumably other aarch64 platforms
https://meta.discourse.org/t/discourse-setup-line-260-0-syntax-error-operand-expected-error-token-is-0/260991

PR #739, opened 2023-10-09
fixup count of physical cores in discourse-setup

BigEd:
For AArch64
See also PR #702 for branch master
Tagging @tgxworld as requested


CvX (approved):
Thank you! (again! 😃)

PR #744, opened 2023-10-16
DEV: Build base images for ARM64 platform in a seperate daily job

tgxworld:
Why this change?

We're currently building a base image for arm64 platforms as part of the
CI workflow. However, building it on Github action via emulation is too
slow and makes it meaningless to include as part of the CI workflow.

What does this change do?

This change moves the building of the bage image targeting the arm64
platform completely to a scheduled job that runs outside of the normal
CI workflow. This will also enable us to build the images for the test
and dev environment eventually as part of the workflow which we intend
to do in a follow up commit.

### DIFF
--- a/.github/workflows/arm64-build.yml
+++ b/.github/workflows/arm64-build.yml
@@ -64,3 +64,50 @@ jobs:
           docker push discourse/base:2.0.$TAG-arm64
           docker push discourse/base:release-arm64
           docker push discourse/base:aarch64
+  test:
+    runs-on: ubuntu-20.04
+    needs: base
+    defaults:
+      run:
+        working-directory: image/discourse_test
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: arm64
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
+      - name: build discourse/discourse_test:slim-arm64
+        run: |
+          docker buildx build . --load \
+            --build-arg from_tag=slim-arm64 \
+            --target base \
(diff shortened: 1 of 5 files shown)


.github/workflows/arm64-build.yml
```diff
@@ -0,0 +1,138 @@
+on:
+  pull_request:
```
tgxworld, line 2: Enabling it on pull requests first so that we run the job in this PR and I know I got things right. Will remove before merging.

.github/workflows/arm64-build.yml
```diff
@@ -0,0 +1,138 @@
...
+    - cron: "0 0 * * *"
+
+concurrency:
+  group: build-arm64-${{ format('{0}-{1}', github.head_ref || github.run_number, github.job) }}
+  cancel-in-progress: true
+
+env:
+  BUILDKIT_PROGRESS: plain
+
+jobs:
```
tgxworld, line 13: There is alot of duplication between this new workflow and `build.yml` but I want to get things running and see if we can even run everything in a reasonable amount of time first.

ZogStriP (approved):

davidtaylorhq:
My preference would be to keep things as-is, so that we continue to get feedback on the ARM build in PRs/commits. If we have a time-sensitive change, we can still choose to merge before it completes.

If we do decide to proceed with this approach, I think we should avoid duplicating the workflow definition, and instead add a conditional to the existing ARM job. We do something similar in plugin/theme workflows [here](https://github.com/discourse/.github/blob/47aa98df7017977d688ffeb529c8fa1c129122d1/.github/workflows/discourse-plugin.yml#L129).

So in this case, to ignore push/pr events, we could do something like:

```yaml
aarch64:
  if: ${{ github.event_name == 'schedule' }}
```

tgxworld:
No longer required because we have access to Github hosted ARM runners now.

PR #745, opened 2023-10-16
DEV: Fix broken ARM64 build

tgxworld:
## Why this change?

The `docker tag discourse/base:aarch64-slim localhost:5000/discourse/base:aarch64-slim
` command broked in 0c93b2207d22b9de69b503460dc269f8b33935c2 because
there is no `aarch64-slim` image being built. Instead, the `ruby
auto_build.rb base_slime_64` command builds an image with the
`discourse/base:build_slim_arm64` tag.

## What does this change do?

This change removes the tagging command because it is pointless for us
to tag it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on:
       - main
   pull_request:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: "0 0 * * *"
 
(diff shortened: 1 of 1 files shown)


lis2 (approved):

tgxworld:
Thank you @lis2

PR #746, opened 2023-10-21
remove unwanted whitespaces

itsbhanusharma:


tgxworld:
Thank you for the contribution but we do not accept cosmetic changes in general 🙇

PR #747, opened 2023-10-23
Discourse wo repo

SoltanYang:
Made it! Code now taken from our github private repository. Docker will build correct image and container is working.


SoltanYang:
> I think you've made a pull request against the wrong repo.

I'm terribly sorry, will fix that.

PR #750, opened 2023-10-27
DEV: Improve support for Ubuntu 20 and Debian 12

Arkshine:
I've been looking to improve the guide to setting up a development environment and tested the different versions of Ubuntu and Debian (fresh dev installation on cloud hosting)

What does this PR is to fix *Ubuntu 20* and *Debian 12* installations.
*Ubuntu 20* is still [officially supported](https://wiki.ubuntu.com/Releases).

In short:
* Removed the Debian backports usage. The `libheif` and `libaom` libraries are not backported.
* `libwebp7` is required in *Debian 12*.

*Debian 10* and *Ubuntu 18* are voluntarily ignored in the changes because they are either no longer supported or too old.

## Debian 

There is no backports installed by default. 
E.g. Debian 11:

```
E: The value 'bullseye-backports' is invalid for APT::Default-Release as such a release is not available in the sources
failed
```

Looking at the Debian backports:
* *Debian 11 Bullseye* https://backports.debian.org/bullseye-backports/overview/
* *Debian 12 Bookworm*: https://backports.debian.org/bookworm-backports/overview/

The libraries are not backported. I don't see any reason to try to use them.
### 10

The script fails on `ImageMagick`. The required version doesn't match.

`checking for libheif >= 1.4.0... no`

```
# apt show libheif1
Package: libheif1
Version: 1.3.2-2~deb10u1
```

It's possible to compile from source. 
This distribution is fairly old, and It seems to be a good idea to use at least the same version as Discourse (*Debian 11*)
### 11

With the change introduced in this PR, it works great.
Discourse uses the `bullseyes-slim` docker image, and the `bullseye-backports` is included by default.
### 12

```
# cat /etc/issue
Debian GNU/Linux 12 \n \l
```
With the change introduced in this PR, it works great.

### 18

Similar to *Debian 10*, this can work with some effort. However, the distribution is fairly old and unsupported.
To make it work, you need at least Node 16 and *libeom*/*heif* compiled from the source.
We can skip it.

### 20
With the change introduced in this PR, it works great.

### 22
It works great as it is.

### DIFF
--- a/image/base/install-imagemagick
+++ b/image/base/install-imagemagick
@@ -8,8 +8,8 @@ IMAGE_MAGICK_HASH="d282117bc6d0e91ad1ad685d096623b96ed8e229f911c891d83277b350ef8
 # We use debian, but GitHub CI is stuck on Ubuntu Bionic, so this must be compatible with both
 LIBJPEGTURBO=$(cat /etc/issue | grep -qi Debian && echo 'libjpeg62-turbo libjpeg62-turbo-dev' || echo 'libjpeg-turbo8 libjpeg-turbo8-dev')
 
-# Ubuntu 22.04/22.10  doesn't have libwebp6
-LIBWEBP=$(cat /etc/issue | grep -qi 'Ubuntu 22' && echo 'libwebp7' || echo 'libwebp6')
+# Ubuntu 22.04/22.10 and Debian 12  doesn't have libwebp6
+LIBWEBP=$(cat /etc/issue | grep -qiE 'Ubuntu 22|Debian.*12' && echo 'libwebp7' || echo 'libwebp6')
 
 PREFIX=/usr/local
 WDIR=/tmp/imagemagick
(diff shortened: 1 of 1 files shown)


Arkshine:
I made an update. I tested Debian 10, 11, and 12, and this PR improves support for Ubuntu 20 and _now_ Debian 12.
The first message should be more descriptive.

Arkshine:
The build failure doesn't seem related to this PR. 🤔  It errors with:

```
#10 5.898 Couldn't connect to Redis
#10 5.930 rake aborted!
#10 5.930 Redis::CannotConnectError: Error connecting to Redis on localhost:6379 (Errno::EADDRNOTAVAIL) (Redis::CannotConnectError)
```

tgxworld:
Thank you for the PR here and this has been addressed in https://github.com/discourse/discourse_docker/commit/21a0351e301d9dde94ae8022ad734827624434f3.

PR #751, opened 2023-10-27
Add tags to pups templates

featheredtoast:
The purpose here is to allow greater flexibility in how and where docker images are built and run. It achieves this by breaking up build steps into distinct run steps which can be saved along the way. Customizable base images may then be prebuilt with as many batteries included as possible, with zero environment setup so those images can then be configured at a later stage.

Add the ability to run partial pups configuration:
`build`: build base image with no db - ember build.
`precompile`: precompile stage that requires postgres and redis.
`migrate`: run migration tasks.
`db`: start bundled postgres/redis, if included.

Adds a create_db script in postgres template for creating db on the fly. Called below in unicorn run:

updates unicorn run command with 3 env flags:
CREATE_DB_ON_BOOT: if 1, creates base db schema, allows for deferral of creation. MIGRATE_ON_BOOT: if 1, runs db:migrate - allows for deferral of db migration. PRECOMPILE_ON_BOOT: if 1, precompiles assets (without ember build).

PRECOMPILE_ON_BOOT initially defaults to 1 in base builds (no tags). During the `precompile` build step, this updates the default to be 0.

All other new flags are default off. With these three flags, we're now able to ship and start a container from a base image, and it'll be able to bootstrap a blank database.

Updates hook to start redis before_db_migrate as before_code hook is not guaranteed to fire before migrate tasks if pups is filtered by tags.

### DIFF
--- a/templates/postgres.template.yml
+++ b/templates/postgres.template.yml
@@ -55,6 +55,7 @@ run:
        sv stop postgres
 
   - file:
+     tag: db
      path: /root/upgrade_postgres
      chmod: "+x"
      contents: |
--- a/templates/postgres.template.yml
+++ b/templates/postgres.template.yml
@@ -17,17 +17,20 @@ hooks:
        to: sv start postgres || exit 1
 
 run:
-  - exec: locale-gen $LANG && update-locale
-  - exec: mkdir -p /shared/postgres_run
-  - exec: chown postgres:postgres /shared/postgres_run
-  - exec: chmod 775 /shared/postgres_run
-  - exec: rm -fr /var/run/postgresql
-  - exec: ln -s /shared/postgres_run /var/run/postgresql
-  - exec: socat /dev/null UNIX-CONNECT:/shared/postgres_run/.s.PGSQL.5432 || exit 0 && echo postgres already running stop container ; exit 1
-  - exec: rm -fr /shared/postgres_run/.s*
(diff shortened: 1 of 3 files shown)


danielwaterworth (approved):
Let's get this merged!

It would be good to put some description of what the tags mean somewhere in the repo. Perhaps in `templates/README.md`.

mentalstring:
We are still on v3.1.5 as we are waiting for some plugins to be compatible with 3.2.

This PR is preventing us from rebuilding the image:

```
rake aborted!
Don't know how to build task 'assets:precompile:build' (See the list of available tasks with `rake --tasks`)
Did you mean?  assets:precompile
               assets:precompile:css
               assets:precompile:before
/var/www/discourse/vendor/bundle/ruby/3.2.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/usr/local/bin/bundle:25:in `load'
/usr/local/bin/bundle:25:in `<main>'
(See full trace by running task with --trace)
I, [2024-04-27T10:45:53.719449 #1]  INFO -- : 
```

Any workaround?

mentalstring:
In case this is useful for someone in the same situation, I managed to rebuild by checking out discourse/discourse_docker to the commit before this one and (for some other unrelated reason, I guess), commenting DISCOURSE_MAXMIND_LICENSE_KEY or else it was failing at another place.

PR #755, opened 2023-11-13
Slim Slim docker image

pierreozoux:
Hi,

we are maintaining an image for kubernetes, and basically, we put together:
 - slim image
 - release image
 - env from web template

But, we don't consider container as slim VMs, but we consider them as "fat" binairies.

As such, we don't use tools like ssh, postgres, redis and so on.

I think in the best world, we'd share a base image that could be suitable for you and us.
But in the mean time, if we could reorganize just a bit this docker image, this would help us follow changes.

Bascially, we don't need this block:
https://github.com/pierreozoux/discourse_docker/blob/patch-1/image/base/slim.Dockerfile#L20-L46

and this block:
https://github.com/pierreozoux/discourse_docker/blob/patch-1/image/base/slim.Dockerfile#L87-L95

Maybe the way we propose is not perfect, but this is to eventually start the discussion?

Thanks for your time and have a nice day!

### DIFF
--- a/image/base/slim.Dockerfile
+++ b/image/base/slim.Dockerfile
@@ -29,7 +29,7 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ bullseye-pgdg main" | \
 # install these without recommends to avoid pulling in e.g. X11 libraries, mailutils
 RUN DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends rsyslog logrotate cron ssh-client less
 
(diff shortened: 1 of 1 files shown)


tgxworld:
Hi @pierreozoux, 

I have a couple of questions:

1. How much savings on the size of the image do you get out of excluding those dependencies?
2. If you really need to remove those dependencies, have you tried building a new image that uninstalls those dependencies?

tgxworld:
@pierreozoux I didn't get a reply previously so I'm closing this off first. Feel free to open a new PR and ping me there if you'll like us to review these changes again.

PR #756, opened 2023-11-14
Bump PostgreSQL major version to 15 from 13

tgxworld:


pfaffman:
The last two installs I did look like they have pg16.

PR #765, opened 2024-01-11
FEATURE: Add early support for aarch64 in dev env

Flink:
This PR adds some new steps to support the aarch64 architecture on Linux.

An updated version of Rust is needed to compile the `selenium-manager` binary as it’s not shipped with the `selenium-webdriver` gem yet.
In the same vein, Google doesn’t ship an aarch64 version of Chrome yet, so it doesn’t make sense to install even Chromium in the image. We have to rely on Firefox to run the system specs.

Related PR: https://github.com/discourse/discourse/pull/25230

### DIFF
--- a/image/base/install-rust
+++ b/image/base/install-rust
@@ -5,7 +5,7 @@ set -e
 export RUSTUP_HOME=/usr/local/rustup
 export CARGO_HOME=/usr/local/cargo
 export PATH=/usr/local/cargo/bin:$PATH
-export RUST_VERSION=1.68.0
+export RUST_VERSION=1.75.0
 export RUSTUP_VERSION=1.25.2
 
 dpkgArch="$(dpkg --print-architecture)"
--- a/image/discourse_dev/Dockerfile
+++ b/image/discourse_dev/Dockerfile
@@ -36,13 +36,13 @@ RUN mv /shared/postgres_data /shared/postgres_data_orig
 # invocations when used with a mounted volume)
 ADD ensure-database /etc/runit/1.d/ensure-database
 
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - &&\
-    echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list &&\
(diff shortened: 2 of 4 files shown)


image/base/install-rust
```diff
@@ -5,7 +5,7 @@ set -e
 export RUSTUP_HOME=/usr/local/rustup
 export CARGO_HOME=/usr/local/cargo
 export PATH=/usr/local/cargo/bin:$PATH
-export RUST_VERSION=1.68.0
+export RUST_VERSION=1.75.0
```
Flink, line 8: Had to update to the latest version of rust to be able to compile `selenium-manager`.

image/discourse_dev/Dockerfile
```diff
@@ -36,13 +36,14 @@ RUN mv /shared/postgres_data /shared/postgres_data_orig
...
+ADD install-selenium /tmp/install-selenium
+RUN /tmp/install-selenium
 
 # Install & Configure MailHog (https://github.com/mailhog/MailHog)
 RUN wget -qO /tmp/mailhog https://github.com/mailhog/MailHog/releases/download/v1.0.1/MailHog_linux_amd64\
     && echo "e2ed634ded49929f089b20045581955ed217672078fd86082dd7a6c67c5d09c7 /tmp/mailhog" | sha256sum -c -\
     && mv /tmp/mailhog /usr/local/bin/mailhog\
     && chmod +x /usr/local/bin/mailhog
+
+RUN rm -rf /tmp/*
```
davidtaylorhq, line 49: We need to do the cleanup in the same step which creates the mess. Cleaning up in a later layer doesn't actually save space in the image.

Best thing might be to move install-rust and all the relevant cleanup in the install-selenium script? That way, it'll just create one layer, and the temp stuff won't be persisted anywhere.

image/discourse_dev/install-selenium
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -e
+
+if [ "$(dpkg --print-architecture)" = "arm64" ]; then
```
davidtaylorhq, line 4: Maybe add a comment to explain why this is different?

davidtaylorhq (approved):

PR #767, opened 2024-01-31
Switch to macOS runner for arm64 variant image building

tgxworld:


tgxworld:
Lol wait we have tried this before.. macOS runners do not support docker.

PR #773, opened 2024-02-29
remove old swapfile script

andrewschleifer:
Today, the official instructions direct self-hosters to the `discourse-setup` which does everything necessary.


mwaniki-wairungu (approved):
LGTM.

PR #779, opened 2024-03-19
Use Github hosted ARM runners to build arm64/aarch64 release image

tgxworld:
### Why this change?

We have been given access to Github's private beta of ARM hosted
runners. Switching to ARM runners should drastically speed up the time
required for us to build our ARM image.

### What does this change do?

1. Switch to use Github's ARM hosted runners.
2. Build release image for arm64 as well. We previously only built the
   slim image because building the release image through emulation is
   way too slow so we skipped the release image.
3. Update `bundle` in `release.Dockerfile` to install gems in parallel
   based on the number of cores instead of hardcoding it to 4 jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,24 +120,45 @@ jobs:
           docker login --username discoursebuild --password $DOCKERHUB_PASSWORD
           docker push discourse/discourse_dev:release
   aarch64:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04${{ ((github.event_name != 'schedule') && '-8core') || '-2core' }}-arm
     needs: base
     steps:
+      - name: Install Docker
+        run: |
+          # Add Docker's official GPG key:
+          sudo apt-get update
+          sudo apt-get install ca-certificates curl
+          sudo install -m 0755 -d /etc/apt/keyrings
+          sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+          sudo chmod a+r /etc/apt/keyrings/docker.asc
+          # Add the repository to Apt sources:
+          echo \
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -120,24 +120,45 @@ jobs:
...
+          sudo apt-get update
+          # Install Docker
+          sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin
+          # Give the current user permission to run docker without sudo
+          sudo usermod -aG docker $USER
+          sudo apt-get install -y acl
+          sudo setfacl --modify user:$USER:rw /var/run/docker.sock
+      - name: Install Ruby
+        run: |
+          sudo apt-get install -y ruby
```
tgxworld, line 148: Note that for the current private beta, Github's ARM hosted runners do not come with any default dependencies installed so we need to install all the dependencies ourselves.

.github/workflows/build.yml
```diff
@@ -120,24 +120,45 @@ jobs:
...
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
```
tgxworld, line 130: No more building through emulation \o/

.github/workflows/build.yml
```diff
@@ -120,24 +120,45 @@ jobs:
...
-      - name: Set up Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v2
```
tgxworld, line 133: This is already installed when we install Docker above.

nattsw (approved):
Hmm… if we have to install stuff, I wonder if it’ll be effectively faster

tgxworld:
>  if we have to install stuff, I wonder if it’ll be effectively faster

We spend about 30seconds to install the dependencies but it is still 999999x faster than building through emulation. 

Screenshot 2024-03-19 at 9 59 38 AM

tgxworld:
Thank you for the review @nattsw 👍

PR #780, opened 2024-03-19
Use Github's ARM hosted runners for building ARM image

tgxworld:
## Why this change?

We have been given access to Github's private beta of ARM hosted
runners. Switching to ARM runners should drastically speed up the time
required for us to build our ARM image.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,36 +15,76 @@ env:
 
 jobs:
   base:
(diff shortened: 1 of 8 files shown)


tgxworld:
Closed by https://github.com/discourse/discourse_docker/commit/fc61b8a850e0a7f8c1ef0852a13d735121b4d862

PR #789, opened 2024-03-27
DEV: Set RUBY_VERSION and DEBIAN_RELEASE env

tgxworld:
## Why this change?

These envs allow the Debian release name and Ruby version to be easily determined without having to know which commands to run.

### DIFF
--- a/image/base/slim.Dockerfile
+++ b/image/base/slim.Dockerfile
@@ -1,13 +1,18 @@
 # NAME:     discourse/base
 # VERSION:  release
-FROM debian:bullseye-slim
 
+ARG DEBIAN_RELEASE=bullseye
+FROM debian:${DEBIAN_RELEASE}-slim
+
+ARG DEBIAN_RELEASE
 ENV PG_MAJOR=13 \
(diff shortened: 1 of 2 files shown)


davidtaylorhq (approved):

image/base/slim.Dockerfile
```diff
@@ -1,13 +1,18 @@
...
+ARG DEBIAN_RELEASE
 ENV PG_MAJOR=13 \
     RUBY_ALLOCATOR=/usr/lib/libjemalloc.so \
     RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH \
-    LEFTHOOK=0
+    LEFTHOOK=0 \
+    RUBY_VERSION=3.2.3 \
+    DEBIAN_RELEASE=${DEBIAN_RELEASE}
```
tgxworld, line 15: Initially I wanted to set `DEBIAN_VERSION` based on the output from `cat /etc/debian_version` but there isn't a way to do so.

Flink (approved):

PR #791, opened 2024-04-15
Launcher2

featheredtoast:
Rewrite of launcher in golang.

Includes one small update to the Redis template, to allow new redis servers to initialize redis_data if it does not exist.

Adds a convenience `launcher2.sh` shellscript which also downloads the binary releases for convenience. We can theoretically replace the existing `launcher` script with this in the future.

The rewritten launcher program also has a method of upgrading, which downloads and installs the latest release binary. It no longer upgrades on run.

For now, I'm aiming to have launcher and launcher2 coexisting side by side. The rewritten launcher2 is aimed at being a drop-in replacement for launcher, with a number of enhancements.

## Changes from launcher

No software prerequisites are checked here. It assumes you have docker set up and whatever minimum requirements setup for Discourse: namely a recent enough version of docker, git.

Some things are not implemented from launcher1.

* `DOCKER_HOST_IP` - container can use `host.docker.internal` in most cases. Supported on mac and windows... can also be [added on linux via docker args](https://stackoverflow.com/questions/72827527/what-is-running-on-host-docker-internal-host).
* debug containers - not implemented. No debug containers saved on build. Under the hood, launcher2 uses docker build which does not allow images to be saved along the way.
* stable `mac-address` - not implemented.

## New features

In a nutshell: split bootstrap/rebuild process up into distinct parts to allow for greater flexibility in how we build and deploy Discourse containers.

### Separates bootstrap process into distinct build, configure, and migrate steps.

Separating the larger bootstrap process into separate steps allows us to break up the work.

`bootstrap` becomes an alias for: `build`, `migrate`, `configure`. There are multiple benefits to this.

#### Build: Easier creation for prebuilt docker images

Share built docker images by only running a `build` step - this build step does not need to connect to a database.
It does not need postgres or redis running. This makes for a simple way to install custom plugins to your Discourse image.

The resulting image is able to be used in Kubernetes and other docker environments.

This is done by deferring finishing the build step, to a later configure step -- which boostraps the db, and precompiles assets.

The `configure` and `migrate` steps can now be done on boot through use of env vars set in the `app.yml` config: `CREATE_DB_ON_BOOT`, `MIGRATE_ON_BOOT`, and `PRECOMPILE_ON_BOOT`, which allows for more portable containers able to drop in and bootstrap themselves and the database as they come into service.

#### Build: Better environment management

The resulting image from a build is a container with no environment (unless `--bake-env` is specified). Additionally, well-known secrets are excluded from the build environment, resulting in a clean history of the prebuilt image that may be more easily shared.

Environment is only bound to a container either with `--bake-env` on build, or on a subsequent `configure` step.

#### Migrate: Adds support to *when* migrations are run

`Build` and `Configure` steps do not run migrations, allowing for external tooling to specify exactly when migrations are run.

`Migrate`, (and`bootstrap`, and `rebuild`) steps are the only ones that run migrations.

#### Migrate: Adds support for *how* migrations are run: `SKIP_POST_DEPLOYMENT_MIGRATIONS` support

the `migrate` step exposes env vars that turn on separate post deploy migration steps.

Allows the ability to turn on and skip post migration steps from launcher when running a stand-alone migrate step.

#### Rebuild: Minimize downtime

Both standalone and multi-container setups' downtime have been minimized for rebuilds

##### Standalone
On standalone builds, only stop the running container after the base build is done.
Standalone sites will only need to be offline during migration and configure steps.

For standalone, `rebuild` runs `build`, `stop`, `migrate`, `configure`, `destroy`, `start`.

##### Multiple container, web only
On multi-container setups or setups with a configured external database using web only containers, rebuilds attempt to run migrations without stopping the container.
A multi-container stays up as migration (skipping post deployment migrations) and as any necessary configuration steps are run. After deploy, post deployment migrations are run to clean up any destructive migrations.

For web-only, `rebuild` runs `build`, `migrate (skip post migrations)`, `configure`, `destroy`, `start`, `migrate`.

#### Rebuild: Serve offline page during downtime

Adds the ability to build and run an image that finishes a build on boot, allowing the server to display an offline page.
For standalone builds above, this allows for the accrued downtime from migration and configure steps to happen more gracefully.

Additional container env vars get turned on by adding the `offline-page.template.yml` template:
  * `CREATE_DB_ON_BOOT`
  * `MIGRATE_ON_BOOT`
  * `PRECOMPILE_ON_BOOT`

These allow containers to boot cleanly from a cold state, and complete db creation, migration, and precompile steps on boot.

During this time, nginx can be up which allows standalone builds to display an offline page.

These variables may also be used for other applications where more flexible bootstrapping is desired.

##### Standalone
On rebuild, a standalone site will skip migration if it detects the presence of `MIGRATE_ON_BOOT` in the app config, and will skip configure steps if it detects the presence of `PRECOMPILE_ON_BOOT` in the app config.

For standalone, `rebuild` runs `build`, `destroy`, `start`, skipping `migrate` and `configure`. The started container then serves an offline page, and runs migrate and precompiles assets before fully entering service.

##### Multiple container, web only
On rebuild, a web only container will act in the same way as a standalone container. This may result in the same downtime as standalone services, as the containers are swapped, and the new container is still responsible for migration and precompiling before serving traffic.

For web-only containers, it may be desired to either ensure that `MIGRATE_ON_BOOT` and `PRECOMPILE_ON_BOOT` are false. Alternatively, you may run with `--full-build` which will ensure that migration and precompile steps are not deferred for the 'live' deploy.

### Multiline env support

Allows the use of multiline env vars so this is valid config, and is passed through to the container as expected:
```
env:
  SECRET_KEY: |
    ---START OF SECRET KEY---
    123456
    78910
    ---END OF SECRET KEY---
```

### More dependable SIGINT/SIGTERM handling.

Launcher wraps docker run commands, which run as children in process trees. Launcher2 does the same, but attempts to kill or stop the underlying docker processes from interrupt signals.

Tools that extend or depend on launcher should be able to send SIGINT/SIGTERM signals to tell launcher to shut down, and launcher should clean up child processes appropriately.

### Docker compose generation.

Allows easier exporting of configuration from discourse's pups configuration to a docker compose configuration.

### Autocomplete support

Run `source <(./launcher2 sh)` to activate completions for the current shell, or add the results of `./launcher2 sh` to your dotfiles

Autocompletes commands, subcommands, and suggests `app` config files from your containers directory. Having a long site name should not feel like a pain to type.

## Maintainability

Golang is well suited as a drop in replacement as just like a shellscript, the deployed binary can still carry minimal assumptions about a particular platform to run. (IE, no dependency on ruby, python, etc)

Golang allows us to use a fully fleshed out programming language to run native yaml parsing: Calling out to ruby through a docker container worked well enough, but got complicated shuffling results through stdout into shell variables.

Launcher has outgrown being a simple wrapper script around Docker. Golang has good support for tests and breaking up code into separate modules to better support further growth around additional subcommands we may wish to add.

## Roadmap

Scaffolding out subcommands, possibly as a later rewrite for `discourse-setup` as having native YAML libraries should make config parsing and editing simpler to do.


featheredtoast:
Updated to only implement `build` command. Full commands still implemented on `launcher2-dev` branch

featheredtoast:
Also if it helps, I've also organized the commits on this ranch to add by feature:
https://github.com/featheredtoast/discourse_docker/commits/launcher2-dev/

The latest commit on that branch are "rest of the misc commands" I was playing with. Was imagining we could support other generation command for various types of exports.

launcher_go/v2/docker/commands.go
```diff
@@ -0,0 +1,54 @@
+package docker
```
tgxworld, line 1: Just curious why we don't use the golang Docker client to execute Docker commands instead?

launcher_go/v2/main.go
```diff
@@ -0,0 +1,65 @@
...
+	"os/exec"
+	"os/signal"
+)
+
+type Cli struct {
+	Version      kong.VersionFlag `help:"Show version."`
+	ConfDir      string           `default:"./containers" help:"Discourse pups config directory." predictor:"dir"`
+	TemplatesDir string           `default:"." help:"Home project directory containing a templates/ directory which in turn contains pups yaml templates." predictor:"dir"`
+	BuildDir     string           `default:"./tmp" help:"Temporary build folder for building images." predictor:"dir"`
+	ForceMkdir   bool             `short:"p" name:"parent-dirs" help:"Create intermediate output directories as required.  If this option is not specified, the full path prefix of each operand must already exist."`
```
tgxworld, line 19: I'm wondering if there is a need to make this configurable. Is it possible for us to just use `/tmp` instead as I feel like these two options can be quite confusing to users.

launcher_go/v2/main.go
```diff
@@ -0,0 +1,65 @@
...
+	"golang.org/x/sys/unix"
+	"os"
+	"os/exec"
+	"os/signal"
+)
+
+type Cli struct {
+	Version      kong.VersionFlag `help:"Show version."`
+	ConfDir      string           `default:"./containers" help:"Discourse pups config directory." predictor:"dir"`
+	TemplatesDir string           `default:"." help:"Home project directory containing a templates/ directory which in turn contains pups yaml templates." predictor:"dir"`
```
tgxworld, line 17: I'm also not sure if we should make this configurable as the help text is only useful for users that understand how `discourse_docker` is structured. I think we should just simplify things and assume that `launcher` has to be ran from the root path of the `discourse_docker` repository.

launcher_go/v2/main.go
```diff
@@ -0,0 +1,65 @@
...
+	BuildDir     string           `default:"./tmp" help:"Temporary build folder for building images." predictor:"dir"`
```
tgxworld, line 18: It seems like we need to add the `./tmp` directory to `.gitignore` if we intent to keep the tmp directory within the repository path.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (config *Config) Yaml() string {
+	return strings.Join(config.rawYaml, "_FILE_SEPERATOR_")
+}
+
+func (config *Config) Dockerfile(pupsArgs string, bakeEnv bool) string {
+	builder := strings.Builder{}
+	builder.WriteString("ARG dockerfile_from_image=" + config.Base_Image + "\n")
+	builder.WriteString("FROM ${dockerfile_from_image}\n")
+	builder.WriteString(config.DockerfileArgs() + "\n")
```
tgxworld, line 137: I am not sure I understand this. The envs being declared in the YAML file is being declared as `ARG`in the Dockerfile. However, I don't see us using the `ARG`s in the Dockerfile which we generate here.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return config, nil
+}
+
+func (config *Config) Yaml() string {
+	return strings.Join(config.rawYaml, "_FILE_SEPERATOR_")
+}
+
+func (config *Config) Dockerfile(pupsArgs string, bakeEnv bool) string {
```
tgxworld, line 133: What is the purpose of generating a Dockerfile on the fly and using it to build an image? Am wondering why we made a choice here to move away from the `cidfile` approach in building a Docker image with pups.

tgxworld: Is this to support a customizable boot command in a sensible way?

launcher_go/v2/docker/commands.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.Dir = r.Dir
+	cmd.Env = r.Config.EnvArray(false)
+	cmd.Env = append(cmd.Env, "BUILDKIT_PROGRESS=plain")
+	for k, _ := range r.Config.Env {
+		cmd.Args = append(cmd.Args, "--build-arg")
+		cmd.Args = append(cmd.Args, k)
+	}
+	cmd.Args = append(cmd.Args, "--no-cache")
+	cmd.Args = append(cmd.Args, "--pull")
+	cmd.Args = append(cmd.Args, "--force-rm")
```
tgxworld, line 40: I see that this is a valid option but couldn't find the documentation for it. Is this used to forcefully stop the build?

launcher_go/v2/docker/commands.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.Args = append(cmd.Args, "-t")
+	cmd.Args = append(cmd.Args, utils.BaseImageName+r.Config.Name+":"+r.ImageTag)
+	cmd.Args = append(cmd.Args, "--shm-size=512m")
```
tgxworld, line 43: curious why we need to increase the shm size here.

launcher_go/v2/config/config.go
tgxworld: O ok I get it now, we want to bake things like `EXPOSE` and `ENV` into the Docker image without having to specify those envs when running the image.

launcher_go/v2/config/config_test.go
```diff
@@ -0,0 +1,39 @@
...
+	AfterEach(func() {
+		os.RemoveAll(testDir)
+	})
+	It("should be able to load", func() {
+		conf, err := config.LoadConfig("../test/containers", "test", true, "../test")
+		Expect(err).To(BeNil())
+		result := conf.Yaml()
+		Expect(string(result)).To(ContainSubstring("DISCOURSE_DEVELOPER_EMAILS: 'me@example.com,you@example.com'"))
+		Expect(string(result)).To(ContainSubstring("_FILE_SEPERATOR_"))
+		Expect(string(result)).To(ContainSubstring("version: tests-passed"))
```
tgxworld, line 28: Since we are dealing with the `Config` struct here, can't we assert against the fields instead of converting it to a string and doing partial match?

launcher_go/v2/config/config_test.go
```diff
@@ -0,0 +1,39 @@
...
+	var testDir string
+	var conf *config.Config
+	BeforeEach(func() {
+		testDir, _ = os.MkdirTemp("", "ddocker-test")
+		conf, _ = config.LoadConfig("../test/containers", "test", true, "../test")
+	})
+	AfterEach(func() {
+		os.RemoveAll(testDir)
+	})
+	It("should be able to load", func() {
```
tgxworld, line 22: I'm not sure what the convention is like but this doesn't indicate to me which function we are testing against in the module.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+		if !includeKnownSecrets && slices.Contains(utils.KnownSecrets, k) {
+			continue
+		}
+		envs = append(envs, k+"="+v)
+	}
+	slices.Sort(envs)
+	return envs
+}
+
+func (config *Config) DockerfileEnvs() string {
```
tgxworld, line 180:
```suggestion
func (config *Config) dockerfileEnvs() string {
```

This seems to be an internal function and does not need to be exported so the function name should be lower case.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+	builder := []string{}
+	for k, _ := range config.Env {
+		builder = append(builder, "ENV "+k+"=${"+k+"}")
+	}
+	slices.Sort(builder)
+	return strings.Join(builder, "\n")
+}
+
+func (config *Config) DockerfileArgs() string {
```
tgxworld, line 189:
```suggestion
func (config *Config) dockerfileArgs() string {
```

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+	builder := []string{}
+	for k, _ := range config.Env {
+		builder = append(builder, "ARG "+k)
+	}
+	slices.Sort(builder)
+	return strings.Join(builder, "\n")
+}
+
+func (config *Config) DockerfileExpose() string {
```
tgxworld, line 198:
```suggestion
func (config *Config) dockerfileExpose() string {
```

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+	"os"
+	"regexp"
+	"runtime"
+	"slices"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+var DefaultBootCommand = "/sbin/boot"
```
tgxworld, line 17:
```suggestion
const DefaultBootCommand = "/sbin/boot"
```

This will not change.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+
+func DefaultBaseImage() string {
```
tgxworld, line 19:
```suggestion
func defaultBaseImage() string {
```

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return "discourse/base:2.0.20231121-0024"
+}
+
+type Config struct {
+	Name            string `yaml:-`
+	rawYaml         []string
+	Base_Image      string            `yaml:,omitempty`
+	Update_Pups     bool              `yaml:,omitempty`
+	Run_Image       string            `yaml:,omitempty`
```
tgxworld, line 31: This doesn't seem to be used anywhere.

launcher_go/v2/config/config.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (config *Config) WriteYamlConfig(dir string) error {
+	file := strings.TrimRight(dir, "/") + "/config.yaml"
+	if err := os.WriteFile(file, []byte(config.Yaml()), 0660); err != nil {
+		return errors.New("error writing config file " + file)
+	}
+	return nil
+}
+
+func (config *Config) BootCommand() string {
```
tgxworld, line 158:
```suggestion
func (config *Config) bootCommand() string {
```

launcher_go/v2/config/config.go
tgxworld, line 133: I think this function should be tested too.

pfaffman: I've been paying only partial attention to this, so I may be off base. When I've done something like what I think this is doing, I built my images with launcher/pups to (and added env variables to migrate database and such on startup).

> What is the purpose of generating a Dockerfile on the fly and using it to build an image? Am wondering why we made a choice here to move away from the cidfile approach in building a Docker image with pups.

I think the reason to use Dockerfile is because it's the industry standard. I've always thought that if this project had started a year or two later, there would be no pups because there would be no reason for it. When it started, docker-compose either didnt' exist, or was pretty unusable. It was years and years before it got included by default.

Software is more useful when more people are using it? Who's using pups other than us?

launcher_go/v2/main.go
featheredtoast: the default values assumes that it's run in the root path... so you're suggesting we remove the option entirely? I defaulted to making any file/directory have sane defaults but be configurable.

launcher_go/v2/config/config.go
featheredtoast: `ARG` similar to `ENV` also gets passed through to the docker's environment, so if your YAML file has:

```
env:
  DISCOURSE_HOSTNAME: 'test.sample.com'
```

that will be set in the dockerfile as `ARG DISCOURSE_HOSTNAME=${DISCOURSE_HOSTNAME}`
on build, this sets the builder's shell environment, (ie: `export DISCOURSE_HOSTNAME=test.sample.com`) and run the docker build.

When building, this will bind the ARG DISCOURSE_HOSTNAME to be `test.sample.com`, and the build will see its environment set up as expected, as `$DISCOURSE_HOSTNAME=test.sample.com` for use during the pups runs.

(Sidenote, I'm passing env var values in the builder's shell environment as this is the only way I've found to pass multiline values.)

launcher_go/v2/docker/commands.go
featheredtoast: It was a legacy option, if running without buildkit:

```
DOCKER_BUILDKIT=0 docker build --help | grep force-rm
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.

      --force-rm                Always remove intermediate containers
```

launcher_go/v2/docker/commands.go
featheredtoast: the shm size is in current launcher and I carried that option over 1-1 here

launcher_go/v2/main.go
featheredtoast: gitignore is updated in one of the future commits I've got lined up

launcher_go/v2/docker/commands.go
featheredtoast: I took a peek at the golang docker client, but it looks quite low level.

I wanted to keep this a relatively 1-1 with launcher v1, and just reaching out to cli (again) was the simplest way ensure that was the case.

Not against rewriting it in the future (there's a few advantages we'd get, like be less dependent on docker cli, maybe inherit other container runtimes, etc) but it wasn't high on the list for first swing here.

launcher_go/v2/config/config.go
featheredtoast: IMO it's better compatibility to be able to convert our pups config out to a dockerfile - we can rely on `docker build` directly for builds, and having a dockerfile available allows easier exporting of our pups configs out to other tools. It feels much more natural to be able to build images using `docker build` imo.

Future commits add exporting of config out to docker-compose, which also allow a user to build via a dockerfile.

launcher_go/v2/docker/commands.go
tgxworld: Got it 👍

launcher_go/v2/main.go
tgxworld: > but be configurable.

Do we have any use cases where this needs to be configurable because I'm not sure if launcher is used anywhere else except for the root path of the repository? My preference now will be to just ship the v1 without making it configurable at all.

launcher_go/v2/main.go
tgxworld: Perhaps we can just make it such that any parent directories we need in `./tmp` is created if it isn't present? I'm not sure if there is value in making this configurable.

launcher_go/v2/main.go
tgxworld: 👍

launcher_go/v2/config/config.go
tgxworld: > this sets the builder's shell environment, (ie: `export DISCOURSE_HOSTNAME=test.sample.com`) and run the docker build.

O this is something new I learnt. I thought the scope of `ARG` is only applied within the `Dockerfile`.

launcher_go/v2/config/config_test.go
featheredtoast: I was writing them as how it made sense to me under `describe Config`. I've updated it more explicitly now.

launcher_go/v2/main.go
featheredtoast: I've updated the option so it's now hidden from CLI help output

launcher_go/v2/main.go
featheredtoast: I've hidden the build dir option and removed the force option (defaults to making parent dirs as needed)

launcher_go/v2/config/config_test.go
featheredtoast: yaml at this point is the concatenated string of all config in yaml form. We didn't need `string(result)` though, so I've updated that.

tgxworld (approved):

tgxworld:
@featheredtoast This looks good to me. When you merge, can you update the commit message to state that this is a WIP?

saulshanabrook:
Quick q from a user here: Is this launcher meant to be used currently or would you recommend users still see the original launcher?

PR #793, opened 2024-04-29
FEATURE: Fix Discourse installation support in China

devhaozi:
This PR does 2 things:

1. Added a new yarn hook to replace the npm mirror before `yarn install`.
2. Modified `web.china.template.yml` to add more mirror sources.

Below is an explanation of these modifications:

- The GitHub proxy added in `web.china.template.yml` has existed in China for many years, and its repository https://github.com/hunshcn/gh-proxy has 6k+ stars, which can ensure its security and stability.
- The NPM mirror site added in `web.china.template.yml` is maintained by Alibaba Group, one of the largest Internet companies in China.
- Modified the Gem mirror in `web.china.template.yml` to the mirror provided by Tsinghua University, one of the top universities in China.
- The reason why sed is used to replace the `yarn.lock` file is because `yarn install --frozen-lockfile` is used for installation below. If the url is not replaced, the NPM mirror will not take effect.

After applying these modifications, I successfully installed Discourse on the Tencent Cloud China server. No more network problems.

### DIFF
--- a/templates/web.china.template.yml
+++ b/templates/web.china.template.yml
@@ -1,10 +1,21 @@
 hooks:
+  before_code:
+    - exec:
+        cmd:
+          - su discourse -c 'git config --global url."https://mirror.ghproxy.com/https://github.com/".insteadOf "https://github.com/"'
+
   before_web:
     - exec:
-       cmd:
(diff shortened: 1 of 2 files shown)


tgxworld:
The changes look pretty safe to me so I'm approving and merging 👍

PR #796, opened 2024-05-03
FEATURE: Update `discourse-setup` to prompt for MaxMind account ID

tgxworld:
In order to download the free MaxMind GeoLite2 databases, an account ID
and license key is required going forward. This commit updates
`discourse-setup` to start prompting the user to provide the MaxMind
Account ID first before asking for the MaxMind license key. If the user
does not provide the Account ID, the script will not prompt for the
license key as we assume the user has opted out.

We are aware that we don't have a reliable way to test for changes to
the `discourse-setup` script but it is what it is at this point in time.
We intend to invest resources in improving things in the future but now
is not the time.

### Review notes

We unfortunately have to test out the changes by hand:

1. Clone this repo
2. Run `sudo SKIP_CONNECTION_TEST=1 ./discourse-setup`
3. Enter random values when prompted because the values don't matter. 
4. The script will print out "Unfortunately, there was an error changing containers/app.yml" but we can ignore it.
5. Run `sudo cat containers/app.yml | grep MAXMIND` to see that the envs have been updated to the values prompted before.

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -354,21 +354,29 @@ read_default() {
   read_default_result=`echo $read_config_result | sed "s/^\([\"']\)\(.*\)\1\$/\2/g"`
 }
 
-assert_maxmind_license_key() {
+assert_maxmind_envs() {
   if  ! grep DISCOURSE_MAXMIND_LICENSE_KEY $web_file >/dev/null 2>&1
   then
-    echo "Adding MAXMIND placeholder to $web_file"
+    echo "Adding MAXMIND_LICENSE_KEY placeholder to $web_file"
     sed -i '/^.*LETSENCRYPT_ACCOUNT_EMAIL.*/a \ \ #DISCOURSE_MAXMIND_LICENSE_KEY: 1234567890123456' $web_file
   fi
-  if  ! grep DISCOURSE_MAXMIND_LICENSE_KEY $web_file >/dev/null 2>&1
+
+  if  ! grep DISCOURSE_MAXMIND_ACCOUNT_ID $web_file >/dev/null 2>&1
+  then
+    echo "Adding MAXMIND_ACCOUNT_ID placeholder to $web_file"
+    sed -i '/^.*LETSENCRYPT_ACCOUNT_EMAIL.*/a \ \ #DISCOURSE_MAXMIND_ACCOUNT_ID: 123456' $web_file
+  fi
+
+
+  if ! grep -e DISCOURSE_MAXMIND_LICENSE_KEY -e DISCOURSE_MAXMIND_ACCOUNT_ID $web_file >/dev/null 2>&1
   then
     cat <<EOF
 
-    Adding DISCOURSE_MAXMIND_LICENSE_KEY to $web_file has failed! This
-    indicates either that your $web_file is very old or otherwise not
-    what the script expects or that there is a bug in this script. The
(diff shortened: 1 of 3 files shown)


tgxworld:
@pfaffman Pinging you for visibility since you added the MAXMIND config 3 years back.

discourse-setup
```diff
@@ -354,17 +354,25 @@ read_default() {
...
+    sed -i '/^.*LETSENCRYPT_ACCOUNT_EMAIL.*/a \ \ #DISCOURSE_MAXMIND_ACCOUNT_ID: 123456' $web_file
+  fi
+
+
+  if ! grep -e DISCOURSE_MAXMIND_LICENSE_KEY -e DISCOURSE_MAXMIND_ACCOUNT_ID $web_file >/dev/null 2>&1
   then
     cat <<EOF
 
-    Adding DISCOURSE_MAXMIND_LICENSE_KEY to $web_file has failed! This
+    Adding DISCOURSE_MAXMIND_ACCOUNT_ID and DISCOURSE_MAXMIND_LICENSE_KEY to $web_file has failed! This
```
nbianca, line 375: Nitpick: I think this line should be 80 characters long like the ones below.

nbianca (approved):

pfaffman:
Hey! Thanks for pinging me @tgxworld ! I've been meaning to take a look at this today but the day got away from me.

Have a look at tests/README.md. It's not real tests, but it's better than nothing. You don't have to run all the stuff by hand, at least, not each one by hand.

tgxworld:
> Have a look at tests/README.md. It's not real tests, but it's better than nothing. You don't have to run all the stuff by hand, at least, not each one by hand.

I think the tests are all broken at this point.

pfaffman:
It could be that something changed in what the script expects since I wrote those tests. If they don't help you, then they don't help you.

It looks like values are getting set, but when it runs a second time to see if it can change already saved values, it's failing. I'm not sure how or why that would be. Ah. Here it is:

In tests/standalone:

-$DIR/../discourse-setup --skip-connection-test --skip-rebuild <<EOF 
+$DIR/../discourse-setup --debug --skip-connection-test --skip-rebuild <<EOF 

In discourse-setup:

   then   
     echo "Stopping existing container in 5 seconds or Control-C to cancel."
     sleep 5  
+    ./launcher stop $app_name                                                   
   else                                                                          
     echo "DEBUG MODE ON. Not stopping the container."
   fi                
-  ./launcher stop $app_name                                                     
   assert_maxmind_license_key                                                    
   assert_notification_email
------

InEnglish: Previously,  the `launcher stop` was not in the if. Also, the test script needs to call `./discourse-setup --debug --skip-connection-tests --skip-rebuild` (--debug was missing). I guess that stuff got added after I last touched the tests?

Adding DEBUG=1 before it runs the second time to change settings would speed things up. But there's still something that keeps the edits from getting processed.

It's a bummer that 1234567890123456 was decided as the placeholder for not having a maxmind key.

You can `check out assert_maxmind_license_key` to see how to add the new required setting to files that don't have them.

I might be inclined to remove support for maxmind from discourse-setup altogether and assume that people who can configure maxmind can manage to use vim/nano to do it by hand.

I hope this helps a bit. 

I'm about to leave town for a couple of weeks and don't have much time to help out. If I did, I'd try to submit a proper PR to fix the test, but I'm horrible at git.

PR #797, opened 2024-05-06
DEV: Run system tests as part of the build

tgxworld:
This commit changes the `base` Github workflow to run system specs as
well. This was previously excluded but it is important for us to run
system specs too since dependecies in the base image can cause system
tests to fail.


Drenmi (approved):
🤖

tgxworld:
The Chrome/Chromium dependency is too complicated for now so going to skip this.

PR #798, opened 2024-05-06
OCSP Stapling added

adityapatadia:


tgxworld:
@adityapatadia Thank you for the contribution :+ I understand how enabling OSCP stapling can be useful but I would like to understand further why you are proposing this as a default for all Discourse instances that uses  `web.letsencrypt.ssl.template.yml`.

templates/web.letsencrypt.ssl.template.yml
```diff
@@ -125,6 +125,10 @@ hooks:
        to: |
          ssl_certificate_key /shared/ssl/$$ENV_DISCOURSE_HOSTNAME.key;
          ssl_certificate_key /shared/ssl/$$ENV_DISCOURSE_HOSTNAME_ecc.key;
+         ssl_stapling on;
+         ssl_stapling_verify on;
```
tgxworld, line 129: According to Nginx's doc,

> For verification to work, the certificate of the server certificate issuer, the root certificate, and all intermediate certificates should be configured as trusted using the [ssl_trusted_certificate](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate) directive.

Are we missing the `ssl_trusted_certificate` directive for this to work?

adityapatadia:
That’s because all users who use docker will benefit from increased speed.
We expect most users will be using lets encrypt and it’s best to have OCSP.

If there is not a good reason to disable it, it should be enabled for all.

Regards,
Aditya Patadia

templates/web.letsencrypt.ssl.template.yml
adityapatadia: It's not required.

andrewschleifer:
Let's Encrypt removed support for OSCP -- https://letsencrypt.org/2024/12/05/ending-ocsp -- in favor of CRLs.

PR #804, opened 2024-05-24
DEV: install redis data folder if none exists

featheredtoast:
With new launcher rewrite, there are instances where redis data folder may not exist on boot. Allow the boot script to create it.

### DIFF
--- a/templates/redis.template.yml
+++ b/templates/redis.template.yml
@@ -86,4 +86,8 @@ hooks:
     - replace:
        filename: /etc/service/unicorn/run
        from: "# redis"
-       to: sv start redis || exit 1
+       to: |
+         if [ ! -d /shared/redis_data ]; then
+           install -d -m 0755 -o redis -g redis /shared/redis_data
+         fi
+         sv start redis || exit 1


tgxworld:
> there are instances where redis data folder may not exist on boot

Can you expand a little on this? Is there something about the launcher that skips 

```
  - exec:
      tag: db
      cmd:
        - install -d -m 0755 -o redis -g redis /shared/redis_data
 ```
 
 if the `tag == db`?

**Comment by featheredtoast** [comment]
Yeah - if the ran tag *excludes* the DB tag, that step would be skipped, this will happen via a `--skip-tags=db` on a pups run.

Idea here is we do not want to assume we have a redis db or data folder on some runs, and this allows us to defer that work until boot.

**Comment by featheredtoast** [comment]
If it helps to expand:

launcher rewrite splits up the classic `bootstrap` into three parts: `build`, `migrate`, `configure`.

Build is everything that can be done without shared folders and database connection. Install plugins, and yarn builds.
Migrate is a db:migrate step.
Configure is the rest of the build that needs to be done with db on and shared folders attached. Compiling theme assets, i18n minification, etc. Basically any setup that needs an environment access to finish the build.

The idea here is to allow certain Discourse containers to be launched to a new, bare environment without migrate or configure steps having been run. On first boot, flags can be given to say "hey go ahead and set yourself up for the first time on this environment" on run, rather than on bootstrap.

For redis, this would mean ensuring the shard drive is created on boot rather than assuming it already exists from a pups run via bootstrap.

File: templates/redis.template.yml [hunk 1]
```diff
@@ -86,4 +86,8 @@ hooks:
     - replace:
        filename: /etc/service/unicorn/run
        from: "# redis"
-       to: sv start redis || exit 1
+       to: |
+         if [ ! -d /shared/redis_data ]; then
+           install -d -m 0755 -o redis -g redis /shared/redis_data
+         fi
```
tgxworld, line 92: @featheredtoast Is there a reason we can't just remove `tag: db` from 

https://github.com/discourse/discourse_docker/blob/6f8c17dbf4def3c991626b51eb9e33ccd92b2296/templates/redis.template.yml#L35-L38

Creating a folder even when we don't need it doesn't seem that bad. I think we should keep the runit service file as simple as possible.

tgxworld:
@featheredtoast This will be solved in https://github.com/discourse/discourse_docker/pull/835/files#diff-92b420ef68e3a6f2f75364f7867a60b754531aff75a9b1c1d546c35c25ae0ad9L36. I think it is fine for us to have an empty directory as long as the Redis template is used. Closing this for now but do let me know if you feel otherwise.

tgxworld:
Reopening because the solution of the other PR has changed.

templates/redis.template.yml
featheredtoast: With the proposed build/configure/migrate separation, we'd be able to push built images that are theoretically able to just stand up with `docker run`, with no additional pups runs or launcher bootstraps, and no preconfigured environment.

It's that case where we'll need to ensure a redis data folder exists on boot.

tgxworld (approved):

PR #816, opened 2024-07-10
DEV: Remove useless lines from `web.template.yml`

tgxworld:
The lines are not necessary because those config has already been set in
when we are building the image.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -191,9 +191,7 @@ run:
       cd: $home
       hook: bundle_exec
       cmd:
-        - su discourse -c 'bundle config --local deployment true'
-        - su discourse -c 'bundle config --local without "development test"'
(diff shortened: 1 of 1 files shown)


templates/web.template.yml
```diff
@@ -191,9 +191,7 @@ run:
       cd: $home
       hook: bundle_exec
       cmd:
-        - su discourse -c 'bundle config --local deployment true'
-        - su discourse -c 'bundle config --local without "development test"'
-        - su discourse -c 'bundle install --retry 3 --jobs 4'
```
tgxworld, line 196: See 

https://github.com/discourse/discourse_docker/blob/dfd3fcee0e2be2b7878923b6c38b362f9b839d1d/image/base/release.Dockerfile#L9-L12

ZogStriP (approved):
🔥

PR #817, opened 2024-07-10
DEV: Updated vanilla.template.yml

jdmartinez1062:
While doing a migrations training following [this meta guide ](https://meta.discourse.org/t/set-up-an-environment-to-migrate-another-forum-to-discourse/88252 ) I found a couple of issues and some things I thought could be improved. 

**Problems and Solutions**

- The usage of `rm -R unicorn` doesn't work anymore as noted [here](https://meta.discourse.org/t/migrate-a-phpbb3-forum-to-discourse/30810/768) due to our migration from Ubuntu to Debian. Because of this, that line of code has been removed which fixed the issue.
- The usage of the old libmariadb package has been updated.
-  The Fetch and reset done to pull custom changes within the Docker instance cause all gems installed in the `vanilla.template.yml` to disappear. Since we reset the gemfile doing so, all installed gems during the import build process get removed (at least from the gemfile) so they don’t work anymore. I updated the `vanilla.template.yml` to execute the fetch and reset before the gems are installed so this doesn't happen anymore. 
- This is not a problem per se but when the `vanila_import.sh` script was run it was also executing the import script. The guide has a step to execute the migration script and recommends the use of `tmux` for longer sessions, but this doesn't allow it. It seemed logical to remove it and allow the user to decide when to execute the migration script

### DIFF
--- a/templates/import/vanilla.template.yml
+++ b/templates/import/vanilla.template.yml
@@ -8,8 +8,9 @@ hooks:
     - exec:
         cd: /etc/service
         cmd:
-          - rm -R nginx
-          - rm -R cron
+          - echo "" > unicorn/run
+          - echo "" > nginx/run
+          - echo "" > cron/run
 
     - exec:
         cd: /etc/runit/3.d
--- a/templates/import/vanilla.template.yml
+++ b/templates/import/vanilla.template.yml
@@ -102,6 +102,13 @@ hooks:
         cmd:
           - mkdir -p /shared/import/data
           - chown discourse -R /shared/import
+  
+  before_code:
+    - exec:
+        cd: $home
+        cmd: 
+          # Add your discourse core fork and pull custom code 
+          - su discourse -c 'git remote set-url origin https://github.com/{github_username}/discourse.git'
 
   after_bundle_exec:
     - exec:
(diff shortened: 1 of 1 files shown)


templates/import/vanilla.template.yml
```diff
@@ -8,9 +8,9 @@ hooks:
     - exec:
         cd: /etc/service
         cmd:
-          - rm -R unicorn
-          - rm -R nginx
-          - rm -R cron
+          - echo "" > unicorn/run
+          - echo "" > nginx/run
+          - echo "" > cron/run
```
gschlager, line 13: Oh, by the way, I think this could actually be replaced with the following snippet:

```yaml
env:
  UNICORN_SIDEKIQS: 0
```

It needs to be at the top of the template and the template needs to be included at after `"templates/web.template.yml"` in the container config. For simplicity, I'd put it at the end of the templates list in the container config.

gschlager (approved):
Looks good! I left you a comment about disabling sidekiq. Sorry that I mentioned this earlier. I completely forgot that there's an easier solution than disabling the services files. :blush:

You can ignore the build error for arm64. That's just a flaky spec in Discourse core.

gschlager (approved):

PR #818, opened 2024-07-12
Fix verification of ports

pabgan:
nc needs -p to publish verification code correctly, at least in Debian 12


SamSaffron:
This looks correct to me cc @pfaffman merging

SamSaffron (approved):

PR #821, opened 2024-08-07
FIX: Remove `bundle config jobs`

tgxworld:
This fixes a regression introduced in
bbefa1e5f387bcad0fac79fea00b39f15f6dee4e. Basically, we cannot configure
the default bundle jobs when building the image because the number of
cores used to build the image can be different from the number of cores
on the machine running the image.


CvX (approved):

tgxworld:
Thank you @CvX 👍

PR #827, opened 2024-08-09
Upgrade nginx to 1.26.1 since it is the new stable version

tgxworld:
Since 1.26.x has been marked stable, 1.25.x is automatically EOL as
nginx only maintains one mainline version and one stable version.


image/base/install-nginx
```diff
@@ -2,13 +2,13 @@
...
-VERSION=1.25.3
-HASH="64c5b975ca287939e828303fa857d22f142b251f17808dfe41733512d9cded86"
+VERSION=1.26.1
 
 cd /tmp
 wget -q https://nginx.org/download/nginx-$VERSION.tar.gz
-sha256sum nginx-$VERSION.tar.gz
-echo "$HASH nginx-$VERSION.tar.gz" | sha256sum -c
+wget -q https://nginx.org/download/nginx-$VERSION.tar.gz.asc
+gpg --import nginx-$VERSION.tar.gz.asc
```
saj, line 10: `gpg --import` will import one or more PGP public keys into the GnuPG key ring.  This `.asc` file is not a public key -- it is a signature.

```
% gpg --import nginx-1.26.1.tar.gz.asc
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
% echo $?
2
```

Release keys are published here.
https://nginx.org/en/pgp_keys.html

```
% curl -Ssf https://nginx.org/keys/nginx_signing.key | gpg --import
gpg: key 2FD21310B49F6B46: 4 signatures not checked due to missing keys
gpg: key 2FD21310B49F6B46: public key "nginx signing key <signing-key-2@nginx.com>" imported
gpg: key ABF5BD827BD9BF62: 6 signatures not checked due to missing keys
gpg: key ABF5BD827BD9BF62: public key "nginx signing key <signing-key@nginx.com>" imported
gpg: key BCDCD8A38D88A2B3: 3 signatures not checked due to missing keys
gpg: key BCDCD8A38D88A2B3: public key "nginx signing key <signing-key-3@nginx.com>" imported
gpg: Total number processed: 3
gpg:               imported: 3
gpg: no ultimately trusted keys found
% echo $?
0
```

The line below this one makes sense.  Once the release keys are imported, you may verify the signature in the `.asc` file.

image/base/pluknet.key
```diff
@@ -0,0 +1,65 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
tgxworld, line 1: See https://github.com/discourse/discourse_docker/pull/827/files#diff-fa8cf8fc82082ca028fdbc192cdfc37c755b819c68eb457c796e3b713e73c5a6R79-R80

image/base/nginx-1.26.1.tar.gz.asc
```diff
@@ -0,0 +1,17 @@
+-----BEGIN PGP SIGNATURE-----
```
tgxworld, line 1: Downloaded from https://nginx.org/en/download.html

image/base/pluknet.key
tgxworld: Obtained from https://nginx.org/en/pgp_keys.html

image/base/install-nginx
tgxworld: Thank you for the tip here 👍 I've applied your feedback into the PR.

image/base/slim.Dockerfile
```diff
@@ -74,6 +74,12 @@ ADD install-jemalloc /tmp/install-jemalloc
 RUN /tmp/install-jemalloc
 
 ADD install-nginx /tmp/install-nginx
+ADD nginx-1.26.1.tar.gz.asc /tmp/nginx-1.26.1.tar.gz.asc
+
+# From https://nginx.org/en/pgp_keys.html and nginx doesn't seem to indicate whose public key was used to sign the package
+# so there is some trial and error required here to figure out the right public key.
+ADD pluknet.key /tmp/nginx_public.key
```
CvX, line 81: So `nginx_signing.key` from that pgp_keys page isn't the key used to sign nginx? 😅

tgxworld: It is apparently used to sign the OS packages and not the files from the download page.

CvX (approved):

image/base/nginx-1.26.1.tar.gz.asc
saj: It is safe to retrieve this signature from the web if you would find that to be more convenient.  (Consider that when this is next bumped, the signature will change.)

Ensuring trust on the key is of sole importance.
With the key vendored, we can use that to verify everything else.
(As a bonus, the key -- or keys -- should not change between releases.)

image/base/slim.Dockerfile
saj: Oh, this confused me as well.

I guess _pluknet_ is Sergey Kandaurov's handle.  Would it help to name the file accordingly?  Perhaps its purpose would be more obvious.

```
image/base/nginx-dev-pluknet.key
```

or

```
image/base/nginx-sergey-kandaurov.key
```

(The file name has no significance to PGP.)

Three other developer keys are listed on that page.  If it is unclear who will be signing future releases, we could vendor and `gpg --import` all four keys.  On `--verify`, `gpg` will pass if it is able to verify a signature against any key.  In other words, this procedure will remain forwards compatible so long as any one of those four people sign a future release.

saj (approved):
LGTM but see comments.

tgxworld:
Thank you @saj and @CvX for reviewing here 🙇

PR #831, opened 2024-08-20
Switch to Debian Bookwork as Bullseye is EOL.

tgxworld:


image/discourse_test/install-chrome
```diff
@@ -10,5 +10,5 @@ else
   wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - &&\
     echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list &&\
     apt update &&\
-    apt install -y google-chrome-stable chromium-driver
```
tgxworld, line 13: chromium is causing all sorts of problem on Debian bookworm so we don't need to install it.

.github/workflows/build.yml
```diff
@@ -104,51 +104,6 @@ jobs:
...
-        if: success() && (github.ref == 'refs/heads/main')
-        env:
-          DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
-        run: |
-          TAG=${{ steps.tag-images.outputs.tag }}
-          docker login --username discoursebuild --password $DOCKERHUB_PASSWORD
-          docker push discourse/base:2.0.$TAG-slim-bookworm
-          docker push discourse/base:slim-bookworm
-          docker push discourse/base:2.0.$TAG-bookworm
-          docker push discourse/base:release-bookworm
```
tgxworld, line 151: This is a temp job I introduced last week so that I can test the Debian Bookworm based image.

.github/workflows/build.yml
```diff
@@ -190,47 +145,6 @@ jobs:
...
-          docker images discourse/discourse_test
-      - name: push to dockerhub
-        if: success() && (github.ref == 'refs/heads/main')
-        env:
-          DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
-        run: |
-          docker login --username discoursebuild --password $DOCKERHUB_PASSWORD
-          docker push discourse/discourse_test:slim-bookworm
-          docker push discourse/discourse_test:slim-browsers-bookworm
-          docker push discourse/discourse_test:release-bookworm
```
tgxworld, line 233: Same as https://github.com/discourse/discourse_docker/pull/831/files#r1722696990

image/discourse_test/install-chrome
davidtaylorhq: What problems were you seeing?

It's still installed in the dev image, and it's kinda essential if we want people to have a decent experience developing on arm64. So just wondering what problems we should look out for?

PR #833, opened 2024-08-20
dev: add user-installed gems to PATH

davidtaylorhq:
This is important for usability of things like discourse_theme, and editor tooling (e.g. ruby-lsp)

### DIFF
--- a/image/discourse_dev/Dockerfile
+++ b/image/discourse_dev/Dockerfile
@@ -12,6 +12,9 @@ ADD sudoers.discourse /etc/sudoers.d/discourse
 
 RUN sudo -u discourse bundle config set --global path /home/discourse/.bundle/gems
 
+# Add user-install ruby gems to PATH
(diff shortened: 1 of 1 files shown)


image/discourse_dev/Dockerfile
```diff
@@ -12,6 +12,9 @@ ADD sudoers.discourse /etc/sudoers.d/discourse
 
 RUN sudo -u discourse bundle config set --global path /home/discourse/.bundle/gems
 
+# Add user-install ruby gems to PATH
+RUN echo 'PATH="$(ruby -r rubygems -e "puts Gem.user_dir")/bin:$PATH"' >>/home/discourse/.profile
```
CvX, line 16: we would usually add whitespace there, right?

```suggestion
RUN echo 'PATH="$(ruby -r rubygems -e "puts Gem.user_dir")/bin:$PATH"' >> /home/discourse/.profile
```

CvX (approved):

PR #842, opened 2024-08-28
FEATURE: add migrate and bootstrap commands

featheredtoast:
Adds command for migrate only. Adds batch commands for bootstrap (build+migrate+configure)

### DIFF
--- a/launcher_go/v2/cli_build.go
+++ b/launcher_go/v2/cli_build.go
@@ -90,6 +90,51 @@ func (r *DockerConfigureCmd) Run(cli *Cli, ctx *context.Context) error {
 	return pups.Run()
 }
 
+type DockerMigrateCmd struct {
+	Config                       string `arg:"" name:"config" help:"config" predictor:"config"`
+	SkipPostDeploymentMigrations bool   `env:"SKIP_POST_DEPLOYMENT_MIGRATIONS" help:"Skip post-deployment migrations. Runs safe migrations only. Defers breaking-change migrations. Make sure you run post-deployment migrations after a full deploy is complete if you use this option."`
+}
+
+func (r *DockerMigrateCmd) Run(cli *Cli, ctx *context.Context) error {
+	config, err := config.LoadConfig(cli.ConfDir, r.Config, true, cli.TemplatesDir)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


launcher_go/v2/cli_build.go
```diff
@@ -90,6 +90,51 @@ func (r *DockerConfigureCmd) Run(cli *Cli, ctx *context.Context) error {
...
+	Config                       string `arg:"" name:"config" help:"config" predictor:"config"`
+	SkipPostDeploymentMigrations bool   `env:"SKIP_POST_DEPLOYMENT_MIGRATIONS" help:"Skip post-deployment migrations. Runs safe migrations only. Defers breaking-change migrations. Make sure you run post-deployment migrations after a full deploy is complete if you use this option."`
+}
+
+func (r *DockerMigrateCmd) Run(cli *Cli, ctx *context.Context) error {
+	config, err := config.LoadConfig(cli.ConfDir, r.Config, true, cli.TemplatesDir)
+	if err != nil {
+		return errors.New("YAML syntax error. Please check your containers/*.yml config files.")
+	}
+	containerId := "discourse-build-" + uuid.NewString()
```
tgxworld, line 103: Let's use a fixed string when running the tests since it makes the `containerId` more deterministic hence more easier to test. Maybe extract what I did into a function?

https://github.com/discourse/discourse_docker/blob/7c18fce1b4cc766be1d3ce5d40edbc9cb8901491/launcher_go/v2/cli_build.go#L71-L79

launcher_go/v2/cli_build.go
```diff
@@ -90,6 +90,51 @@ func (r *DockerConfigureCmd) Run(cli *Cli, ctx *context.Context) error {
...
+	env := []string{"SKIP_EMBER_CLI_COMPILE=1"}
```
tgxworld, line 104: I'm curious why this is needed when we are only running stuff tagged `db,migrate`.

tgxworld (approved):

tgxworld:
Merging first, I'll follow up with a PR to address my review comments.

launcher_go/v2/cli_build.go
featheredtoast: ahh this one's probably not necessary come to think of it, believe you're correct here 👍

PR #847, opened 2024-08-29
FEATURE: add github workflows

featheredtoast:
Add github workflows for CI lint, tests

Also adds support for releases:

Job to build latest version and uploads on successful lint+test, along with a note of the current version.

Job to build on new release creation, in case we want to create stable releases in the future here.

### DIFF
--- a/.github/workflows/launcher_go.yml
+++ b/.github/workflows/launcher_go.yml
@@ -1,0 +1,106 @@
+name: launcher_go
+
+on:
+  release:
+    types: [created]
+  push:
+    branches:
+      - main
+  pull_request: {}
+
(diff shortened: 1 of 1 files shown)


featheredtoast:
tests are only failing due to docker paths not being what we expect but it is running and it does look good otherwise

tgxworld:
> tests are only failing due to docker paths

We will need to fix this though before we can merge this. Otherwise, we won't be able to build new base images.

Also, I think we need to only run the workflow when files in `launcher2` changes. 

Maybe you can push to a branch on this repo instead so I can have a look to see if I can apply the changes?

PR #849, opened 2024-08-30
FEATURE: add runtime features

featheredtoast:
adds commands for: start, run, stop, cleanup, destroy, logs, enter, restart, rebuild -- carrying over existing run commands from launcher1.

Rebuild will also do its best to minimize downtime with the following steps:

* Detect if Discourse is running as a single container or external DB
* Detect if db:migrate is configured to run on container boot
* Build initial container (keeping existing one online)
* Exit running containers if it's a single container (otherwise keeps existing online)
* Run migrations
  * Defer migrations if db:migrate is configured to run on container boot
  * Run migrations with SKIP_POST_DEPLOYMENT_MIGRATIONS=1 if it's a 2 container setup
  * Otherwise, run all migrations
* Destroy the old container (finally stopping the current, if it's still up here)
* Start the new container
* Run post-deploy migrations
  * Run migrations with SKIP_POST_DEPLOYMENT_MIGRATIONS=0 if it's a 2 container setup

### DIFF
--- a/launcher_go/v2/cli_runtime_test.go
+++ b/launcher_go/v2/cli_runtime_test.go
@@ -1,0 +1,191 @@
+package main_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"bytes"
+	"context"
+	ddocker "github.com/discourse/discourse_docker/launcher_go/v2"
+	. "github.com/discourse/discourse_docker/launcher_go/v2/test_utils"
+	"github.com/discourse/discourse_docker/launcher_go/v2/utils"
+	"os"
+)
+
+var _ = Describe("Runtime", func() {
+	var testDir string
+	var out *bytes.Buffer
+	var cli *ddocker.Cli
+	var ctx context.Context
+
+	BeforeEach(func() {
+		utils.DockerPath = "docker"
+		out = &bytes.Buffer{}
+		utils.Out = out
+		testDir, _ = os.MkdirTemp("", "ddocker-test")
+
+		ctx = context.Background()
+
+		cli = &ddocker.Cli{
+			ConfDir:      "./test/containers",
+			TemplatesDir: "./test",
+			BuildDir:     testDir,
+		}
+		utils.CmdRunner = CreateNewFakeCmdRunner()
+	})
+	AfterEach(func() {
+		os.RemoveAll(testDir)
+	})
+
+	Context("When running run commands", func() {
+
+		var checkStartCmd = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -q --filter name=test"))
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -a -q --filter name=test"))
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker run"))
+			Expect(cmd.String()).To(ContainSubstring("-d"))
+			Expect(cmd.String()).To(ContainSubstring("--name test local_discourse/test /sbin/boot"))
+		}
+
+		var checkStartCmdWhenStarted = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -q --filter name=test"))
+		}
+
+		var checkStopCmd = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -a -q --filter name=test"))
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker stop -t 600 test"))
+		}
+
+		var checkStopCmdWhenMissing = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -a -q --filter name=test"))
+		}
+
+		It("should run start commands", func() {
+			runner := ddocker.StartCmd{Config: "test"}
+			runner.Run(cli, &ctx)
+			checkStartCmd()
+		})
+
+		It("should not run stop commands", func() {
+			runner := ddocker.StopCmd{Config: "test"}
+			runner.Run(cli, &ctx)
+			checkStopCmdWhenMissing()
+		})
+
+		Context("with a running container", func() {
+			BeforeEach(func() {
+				//response should be non-empty, indicating a running container
+				response := []byte{123}
+				CmdOutputResponse = response
+			})
+
+			It("should not run start commands", func() {
+				runner := ddocker.StartCmd{Config: "test"}
+				runner.Run(cli, &ctx)
+				checkStartCmdWhenStarted()
+			})
+
+			It("should run stop commands", func() {
+				runner := ddocker.StopCmd{Config: "test"}
+				runner.Run(cli, &ctx)
+				checkStopCmd()
+			})
+
+			It("should keep running during commits, and be post-deploy migration aware when using a web only container", func() {
+				runner := ddocker.RebuildCmd{Config: "web_only"}
+
+				runner.Run(cli, &ctx)
+
+				//initial build
+				cmd := GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker build"))
+
+				//migrate, skipping post deployment migrations
+				cmd = GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker run"))
+				Expect(cmd.String()).To(ContainSubstring("--tags=db,migrate"))
+				Expect(cmd.String()).To(ContainSubstring("--env SKIP_POST_DEPLOYMENT_MIGRATIONS=1"))
+
+				// precompile
+				cmd = GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker run"))
+				Expect(cmd.String()).To(ContainSubstring("--tags=db,precompile"))
+				cmd = GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker commit"))
+				cmd = GetLastCommand()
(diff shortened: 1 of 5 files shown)


launcher_go/v2/cli_runtime_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+		var checkStartCmd = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -q --filter name=test"))
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -a -q --filter name=test"))
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker run"))
+			Expect(cmd.String()).To(ContainSubstring("-d"))
+			Expect(cmd.String()).To(ContainSubstring("--name test local_discourse/test /sbin/boot"))
```
tgxworld, line 50: I printed out the full command here and noticed that we are still passing `--ENV` through to `docker run`. My impression is that this is not necessary because the `ENV`s are already baked into the image?

launcher_go/v2/cli_runtime_test.go
```diff
@@ -0,0 +1,191 @@
...
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker stop -t 600 test"))
+		}
+
+		var checkStopCmdWhenMissing = func() {
+			cmd := GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps -a -q --filter name=test"))
+		}
+
+		It("should run start commands", func() {
```
tgxworld, line 70:
```suggestion
		It("should run start commands when the container has not been started", func() {
```

launcher_go/v2/cli_runtime_test.go
```diff
@@ -0,0 +1,191 @@
...
+			runner := ddocker.StartCmd{Config: "test"}
+			runner.Run(cli, &ctx)
+			checkStartCmd()
+		})
+
+		It("should not run stop commands", func() {
```
tgxworld, line 76:
```suggestion
		It("should not run stop commands when the container has not been started", func() {
```

launcher_go/v2/cli_runtime_test.go
```diff
@@ -0,0 +1,197 @@
...
+			Expect(cmd.String()).To(ContainSubstring("docker ps --quiet --filter name=test"))
+
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker ps --all --quiet --filter name=test"))
+
+			cmd = GetLastCommand()
+			Expect(cmd.String()).To(ContainSubstring("docker run"))
+			Expect(cmd.String()).To(ContainSubstring("--detach"))
+			Expect(cmd.String()).To(ContainSubstring("--restart=always"))
+			Expect(cmd.String()).To(ContainSubstring("--name test local_discourse/test /sbin/boot"))
```
tgxworld, line 54: I printed out the full command here and noticed that we are still passing --ENV through to docker run. My impression is that this is not necessary because the ENVs are already baked into the image?

launcher_go/v2/cli_runtime_test.go
```diff
@@ -0,0 +1,204 @@
...
+				// stop
+				Expect(cmd.String()).To(ContainSubstring("docker ps --all --quiet --filter name=standalone"))
+				cmd = GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker stop"))
+
+				// run migrate
+				cmd = GetLastCommand()
+				Expect(cmd.String()).To(ContainSubstring("docker run"))
+				Expect(cmd.String()).To(ContainSubstring("--tags=db,migrate"))
+				Expect(cmd.String()).ToNot(ContainSubstring("--env SKIP_POST_DEPLOYMENT_MIGRATIONS=1"))
```
tgxworld, line 177: It seems like we are skipping post deploy migrations here but we don't seem to be running the migrations after.

launcher_go/v2/cli_runtime_test.go
featheredtoast: This is a bit of a chance as passing env vars are needed for when you're running images without env baked in with a `configure` step -- IE, I would like to be able to run images that have only undergone a `build` step, but still need env from the yaml configuration.

In my mind, the env is baked in here (on configure) really only to allow the image to be deployed out to other sites not run under launcher.

launcher_go/v2/cli_runtime_test.go
featheredtoast: This test is to check standalone bootstraps, which does not run skip post deploy migrations, so no need to run migrations at the end.

Maybe easily missed, but the expects is `ToNot` rather than `To` here

launcher_go/v2/cli_runtime_test.go
tgxworld: > I would like to be able to run images that have only undergone a build step

is there an example you can provide here which demonstrates this use case and its benefits? From what I understand now, we don’t run migration and don’t pre compile assets at the build step. So an image that has been built is pretty much not runnable.

featheredtoast: Of course - I'm trying to allow images to be able to finish bootstrapping via an environment variable switch.

For standalone images, it allows for rebuilds to finish up compiling (non-js) assets and database migrations while having the new image up and running, serving requests -- During this time we can have nginx running and serving an offline page as the container finishes up.

It also allows us to have a stock-standard docker image where someone can just take a normal image, built from a `build` step with zero configuration and be able to `docker run` a discourse and have it run, with no launcher required.

tgxworld: > For standalone images, it allows for rebuilds to finish up compiling (non-js) assets and database migrations while having the new image up and running, serving requests -- During this time we can have nginx running and serving an offline page as the container finishes up.

Even if we want to allow this, wouldn't the user still be running `launcher2 rebuild` which means `build` -> `migrate` -> `configure`. I still don't understand why we need to pass the envs to `launcher2 run`.

> It also allows us to have a stock-standard docker image where someone can just take a normal image, built from a build step with zero configuration and be able to docker run a discourse and have it run, with no launcher required.

How does this tie back to `launcher2 run` passing in ENVs through `--ENV` if `launcher` is not required?

launcher_go/v2/cli_runtime_test.go
tgxworld: > Maybe easily missed, but the expects is ToNot rather than To here

Ahh good catch 👍

launcher_go/v2/cli_runtime_test.go
featheredtoast: The idea is for standalone container setups, the new container can finish up the configure and migrate steps while serving traffic to an offline page.

There are two new env vars here that launcher can take advantage of:
`PRECOMPILE_ON_BOOT` -- which directs the container to run `SKIP_EMBER_CLI_COMPILE=1 su discourse -c 'bundle exec rake assets:precompile'` on boot

`rebuild` can take advantage of these by skipping the `configure` step.

Further, in combination with `MIGRATE_ON_BOOT` -- which directs the container to run `su discourse -c 'bundle exec rake db:migrate'` on boot we can serve an offline page with a single container. The result of a `launcher2 rebuild` for a standalone with the two above env vars set is:

`build` - new image is built (old image stays up)
`stop` - old image stops (container downtime begins)
`start` - new container starts. nginx starts serving offline page (container downtime ends). Since the container is configured with the above env vars, it finished migration + configuration stages on boot. This skips the `launcher2 configure` step because the container is able to do it all on boot.

Without the above configuration, `launcher2 rebuild` runs as it does today:

`build` - builds new image (old image stays up)
`stop` - stops old image (container downtime starts here)
`migrate` - runs `db:migrate`
`configure` - takes new build image and runs assets:precompile. binds env to image.
`start` - starts new image (container downtime ends)

featheredtoast: See also the offline-page template, which configures nginx to serve the error page, and sets the variables for the above.

https://github.com/discourse/discourse_docker/blob/main/templates/offline-page.template.yml

tgxworld: > `build` - builds new image (old container stays up)
> `stop` - stops old container (container downtime starts)
> `migrate` - runs `db:migrate`
> `configure` - takes new build image and runs assets:precompile. binds env to image.
> `start` - starts new container (container downtime ends)

Something doesn't feel right to me here. Why do we stop the old container while we run migrations and assets precompilation? The old container can definitely continue running as long as we skip post deploy migrations. I think we should have no problems doing something like the following? 

1. `build`
2. `migrate`
3. `configure`
4. `stop`
5. `start`

featheredtoast: Launcher2 does exactly that when we detect the user is running external DBs (IE, when we are running 2 container setups).

For standalone setups, we don't have the luxury - we have to ensure we do not have two instances of the image running at the same time because we'd have two containers running postgres, both mounting the same postgres data from the shared folder.

In the postgres template we check the shared folder to see if another container is running here: https://github.com/discourse/discourse_docker/blob/main/templates/postgres.template.yml#L60 (the same checks are also needed for launcher today).

tgxworld: > we have to ensure we do not have two instances of the image running at the same time because we'd have two containers running postgres, both mounting the same postgres data from the shared folder.

The second container doesn't have to start postgres to run migration though. We can just mount the socket which will allow us to connect to PG. I'll try and see if I can change this in the future.

tgxworld (approved):

launcher_go/v2/cli_runtime_test.go
featheredtoast: ooh I didn't know that was even an option -- that would be *fantastic* to see!

PR #852, opened 2024-08-30
DEV: remove bundle cache

featheredtoast:
On the current image this removes ~78MB from the layer

Uncompressed image sizes:
before: 3.3GB
after: 3.23GB


tgxworld (approved):

tgxworld (changes requested):
Looks like the build is not happy 🤔

featheredtoast:
aw phoo, seems like this is required to keep then

featheredtoast:
I've updated the find + delete, looks like there's a cache folder within */gems/* as well -- added to the ignore list on the find command. 👍

tgxworld (approved):

PR #853, opened 2024-08-30
Switch to `discourse/ruby:3.3.4-bookworm-slim`

tgxworld:
This commit switches us to use `discourse/ruby:3.3.4-bookworm-slim` which is build and pushed by the `build` workflow in the https://github.com/discourse/docker-ruby repository. The benefits of doing so is that we reduce the amount of time spent building this image since building Ruby involves installing Rust and compiling it from scratch. Also, we centralise on how Ruby is installed at Discourse making it easy to update Ruby when there are security issues or bugs. 

#### Before
[9m 2s](https://github.com/discourse/discourse_docker/actions/runs/10641111556/job/29501839919)

#### After
[5m 36s](https://github.com/discourse/discourse_docker/actions/runs/10634563252/job/29482161907?pr=853)


image/base/install-oxipng
```diff
@@ -2,23 +2,24 @@
...
-
-cd /tmp
-wget -q https://github.com/shssoichiro/oxipng/archive/refs/tags/${OXIPNG_ARCHIVE}
-sha256sum ${OXIPNG_ARCHIVE}
-echo "${OXIPNG_HASH} ${OXIPNG_ARCHIVE}" | sha256sum -c
+case "${dpkgArch##*-}" in
+    amd64) OXIPNG_FILE="oxipng-${OXIPNG_VERSION}-x86_64-unknown-linux-musl.tar.gz"; OXIPNG_HASH='211d53f3781be4a71566fbaad6611a3da018ac9b22d500651b091c2b42ebe318' ;;
+    arm64) OXIPNG_FILE="oxipng-${OXIPNG_VERSION}-aarch64-unknown-linux-musl.tar.gz"; OXIPNG_HASH='818d47d7195e1e0c4d58a9f3b6fd84aa3cd21770c60c876e73e2e6a17ca69b52' ;;
+    *) echo >&2 "unsupported architecture: ${dpkgArch}"; exit 1 ;;
+esac
```
tgxworld, line 12: OxiPNG started shipping releases for linux arm64 so we no longer have to compile this thus removing the need for us to install Rust in the base image.

PR #858, opened 2024-09-03
FEATURE: add generate docker compose

featheredtoast:
Adds the ability to generate a docker-compose setup from a discourse pups config.

This is an optional feature, and not required for launcher2's runtime lifecycle. If it's deemed launcher shouldn't be responsible for this, I can probably figure out a good way to do this outside the project separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ image/discourse_dev/postgres.template.yml
 image/discourse_dev/redis.template.yml
 .gc-state/*
 .vagrant/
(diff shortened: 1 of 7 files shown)


tgxworld:
> If it's deemed launcher shouldn't be responsible for this,

I feel like we should leave this out until we have plans/needs to rely on `docker-compose`.

featheredtoast:
thought as much, we'll carry on without it for now 👍

PR #864, opened 2024-09-06
Release `discourse/base` images for `discourse/discourse` stable branch

tgxworld:
This commit updates our `build` workflow to push the following manifests
which can be used to run Discourse against the stable branch.

1. discourse/base:2.0.$TIMESTAMP-stable
2. discourse/base:release-stable


image/base/release.Dockerfile
```diff
@@ -1,15 +0,0 @@
...
-ENV RAILS_ENV=production
-
-RUN cd /var/www/discourse &&\
-    sudo -u discourse bundle config --local deployment true &&\
-    sudo -u discourse bundle config --local path ./vendor/bundle &&\
-    sudo -u discourse bundle config --local without test development &&\
-    sudo -u discourse bundle install --jobs $(($(nproc) - 1)) &&\
-    sudo -u discourse /bin/bash -c 'if [ -f yarn.lock ]; then yarn install --frozen-lockfile && yarn cache clean; else pnpm install --frozen-lockfile; fi' &&\
-    find /var/www/discourse/vendor/bundle -name cache -not -path '*/gems/*' -type d -exec rm -rf {} + &&\
-    find /var/www/discourse/vendor/bundle -name tmp -type d -exec rm -rf {} +
```
tgxworld, line 15: Switch to multi-stage Docker builds instead which makes the Dockerfile much easier to understand IMO.

.github/workflows/build.yml
```diff
@@ -43,54 +43,112 @@ jobs:
...
+      - name: Build slim image `--target discourse_dependencies`
+        run: |
+          cd image && ruby auto_build.rb base_slim_${{ matrix.arch }} --cache-to=type=local,dest=/tmp/build_deps_${{ matrix.arch }}
+
+      - name: Print cache
+        run: |
+          ls -lah /tmp/build_deps_${{ matrix.arch }}
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
```
davidtaylorhq, line 63: IIRC we have quite a restrictive limit on artifact storage, so I'm not sure this use of artifacts to move things between build stages is acceptable 😢 

So I think it's probably best to keep everything in  a single job

image/base/Dockerfile
```diff
@@ -127,7 +127,22 @@ RUN rm -f /etc/service
 COPY etc/  /etc
 COPY sbin/ /sbin
 
+FROM discourse_dependencies AS discourse_slim
+ARG DISCOURSE_BRANCH=main
+
```
davidtaylorhq, line 132: How about we just drop the discourse/discourse clone from the slim image? I don't think it's serving any purpose there.

That way, we can avoid the need for the extra complexity of the discourse_dependencies image.

image/base/Dockerfile
```diff
@@ -127,7 +127,22 @@ RUN rm -f /etc/service
...
+ENV RAILS_ENV=production
+
+RUN cd /var/www/discourse &&\
+    sudo -u discourse bundle config --local deployment true &&\
+    sudo -u discourse bundle config --local path ./vendor/bundle &&\
+    sudo -u discourse bundle config --local without test development &&\
+    sudo -u discourse bundle install --jobs $(($(nproc) - 1)) &&\
+    sudo -u discourse /bin/bash -c 'if [ -f yarn.lock ]; then yarn install --frozen-lockfile && yarn cache clean; else pnpm install --frozen-lockfile; fi' &&\
+    find /var/www/discourse/vendor/bundle -name cache -not -path '*/gems/*' -type d -exec rm -rf {} + &&\
+    find /var/www/discourse/vendor/bundle -name tmp -type d -exec rm -rf {} +
```
davidtaylorhq, line 148: How about splitting these up into separate Ruby / JS layers? It'll improve parallelisation of pulls/push, and it'll make it easier for us to see the relative resource usage of each.

.github/workflows/build.yml
```diff
@@ -43,54 +43,112 @@ jobs:
...
       - name: build slim image
         run: |
-          cd image && ruby auto_build.rb base_slim_$ARCH
+          cd image && ruby auto_build.rb base_slim_${{ matrix.discourse_branch }}_${{ matrix.arch }} --cache-from=type=local,src=/tmp/build_deps_${{ matrix.arch }}
+
       - name: tag slim images
         run: |
-          docker tag discourse/base:build_slim_$ARCH discourse/base:2.0.$TIMESTAMP-slim-$ARCH
-          docker tag discourse/base:build_slim_$ARCH discourse/base:slim-$ARCH
+          docker tag discourse/base:build_slim_${{ matrix.discourse_branch }}_${{ matrix.arch }} discourse/base:2.0.${{ env.TIMESTAMP }}-slim-${{ matrix.discourse_branch }}-${{ matrix.arch }}
```
davidtaylorhq, line 106: Why all the changes from `$ARCH` to `${{ matrix.arch }}`? IMO the more-concise environment variable syntax is much easier to read/write?

tgxworld: O $ARCH is easier to read/write, it makes debugging stuff in GitHub’s output harder because the command printed is without the ENV substitution.

.github/workflows/build.yml
```diff
@@ -43,54 +43,112 @@ jobs:
...
-          cd image && ruby auto_build.rb discourse_test_build_$ARCH
+          cd image && ruby auto_build.rb discourse_test_build_${{ matrix.discourse_branch }}_${{ matrix.arch }}
+
       - name: run specs
         run: |
-          docker run --rm -e RUBY_ONLY=1 -e USE_TURBO=1 -e SKIP_PLUGINS=1 -e SKIP_LINT=1 discourse/discourse_test:build_$ARCH
-      - name: build & tag dev image
+          docker run --rm -e RUBY_ONLY=1 -e USE_TURBO=1 -e SKIP_PLUGINS=1 -e SKIP_LINT=1 discourse/discourse_test:build_${{ matrix.discourse_branch }}_${{ matrix.arch }}
+
+      - name: build & tag dev image for main Discourse branch
```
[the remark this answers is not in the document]
tgxworld: I did think about that too but it turns out that we have always been cloning main only so I wanted to keep the status-quo.

image/base/Dockerfile
tgxworld: Yup I can do this.

.github/workflows/build.yml
tgxworld: I think I’m overthinking here, ARM and AMD are already built separately from scratch so the risk is already there.

davidtaylorhq: Yeah... agree it's not ideal, but we do already have that risk with arm/amd.

Looks like [the free storage limit](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) is 2GB across our entire organisation 😢

image/base/Dockerfile
davidtaylorhq: AFAIK the only thing which uses base:slim is the discourse_test build (that was the original motivation for us introducing the slim image - to reduce the test image size for github actions).

So I think we can get away with removing the clone from slim, and updating discourse_test to make its own clone

.github/workflows/build.yml
```diff
@@ -152,13 +193,13 @@ jobs:
       - name: build discourse_test:slim
         run: |
           docker buildx build . --load \
-            --build-arg from_tag=slim \
```
tgxworld, line 155: The `slim` tag for `discourse/base` no longer exists. We need to be using `release-slim`.

davidtaylorhq: This was a mistake when I refactored the build recently for multiarch support. Shall we go back to calling it `:slim`? Or do you prefer `release-slim`? (I don't mind either way 🤷‍♂️)

tgxworld: Maybe slim just for consistency sake since that has always been the case?

templates/web.template.yml
```diff
@@ -182,7 +181,7 @@ run:
       hook: yarn
       cmd:
         - |-
-          if [ "$version" != "tests-passed" ]; then
+          if [ "$version" != "tests-passed" ] && [ "$clean_up_node_modules" == true ]; then
```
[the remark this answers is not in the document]
davidtaylorhq: I'll push a commit with updated logic

.github/workflows/build.yml
davidtaylorhq: 💯

templates/web.template.yml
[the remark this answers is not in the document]
tgxworld: Does this mean nothing is broken if stable had been working with this logic?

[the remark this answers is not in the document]
davidtaylorhq: We originally added this to workaround an edge-case with a single dependency in yarn. I guess that edge-case no longer exists, which is why this hasn't been causing a problem.

Now, we have a new problem with the change from yarn -> pnpm. This one is not an edge-case - it's 100% guaranteed to blow up. But, it hasn't affected anyone yet, because the current launcher base image is still yarn-based.

Here's a PR with improved logic: https://github.com/discourse/discourse_docker/pull/869

tgxworld:
Thank you @davidtaylorhq

PR #865, opened 2024-09-06
Clean up nginx folder after installation.

tgxworld:


image/base/install-nginx
```diff
@@ -30,7 +30,7 @@ make install
 mv /usr/share/nginx/sbin/nginx /usr/sbin
 
 cd /
-rm -fr /tmp/nginx
```
tgxworld, line 33: There is no `/tmp/nginx` folder

nbianca (approved):

PR #871, opened 2024-09-19
Add nginx config file and support for outlets

nbianca:
The nginx config file used to be copied from the discourse/discourse repository, but it has been now moved in this project, closer to the place where it is used.

The config has several 'include' statements that implement support for outlets that templates can then use to extend the default configuration for various features. This is an alternative to the "find & replace" hacks.


templates/web.template.yml
```diff
@@ -203,24 +191,6 @@ run:
...
-      tag: build
-      hook: assets_precompile_build
-      cmd:
-        - su discourse -c 'bundle exec rake assets:precompile:build'
-  - exec:
-      cd: $home
-      tag: precompile
-      hook: assets_precompile
-      cmd:
-        - su discourse -c 'SKIP_EMBER_CLI_COMPILE=1 bundle exec rake themes:update assets:precompile'
```
tgxworld, line 223: Just want to make sure that we are not removing these intentioanlly?

templates/web.ssl.template.yml
```diff
@@ -1,56 +1,47 @@
...
-       location @discourse {
-       add_header Strict-Transport-Security 'max-age=31536000'; # remember the certificate for a year and automatically connect to HTTPS for this domain
+      cmd:
+        - "mkdir -p /shared/ssl/"
+
+  - file:
+      path: "/etc/nginx/conf.d/outlets/before-server/redirect-http-to-https.conf"
+      contents: |
+        server {
+          include conf.d/outlets/http/*.conf;
```
tgxworld, line 10: Can this outlet be moved into the base template? I don't think we should be injecting outlets in pups cause it will be hard to find.

tgxworld: Oops sorry, I misread what we are doing here.

templates/web.template.yml
featheredtoast: this doesn't look like it should be removed from the template, my guess was this was temporarily removed for testing the nginx template updates @nbianca ?

nbianca: Oh, that's right. My bad. I will add them back. :)

launcher
```diff
@@ -693,7 +693,7 @@ run_bootstrap() {
   declare -i BOOTSTRAP_EXITCODE
   rm -f $cidbootstrap
 
-  echo "$input" | $docker_path run --shm-size=512m $user_args $links "${env[@]}" -e DOCKER_HOST_IP="$docker_ip" --cidfile "$cidbootstrap" -i -a stdin -a stdout -a stderr $volumes $image \
+  echo "$input" | $docker_path run --shm-size=512m $user_args $links "${env[@]}" -e DOCKER_HOST_IP="$docker_ip" --cidfile "$cidbootstrap" -i -a stdin -a stdout -a stderr -v $PWD:/discourse_docker $volumes $image \
```
tgxworld, line 696: We can just include the templates under `/etc/nginx/templates/*.template` instead of having to mount a new volume when running the image. The files are not huge so the increase in disk space should not have a significant impact on the overall image size.

templates/web.ipv6.template.yml
```diff
@@ -1,8 +0,0 @@
-run:
```
tgxworld, line 1: I think we still need to keep this template around but make it a noop? Maybe echo a deprecation message to state that this will be removed by a given date?

templates/web.ssl.template.yml
```diff
@@ -1,56 +1,52 @@
...
+      path: "/etc/nginx/conf.d/outlets/before-server/redirect-http-to-https.conf"
+      contents: |
+        server {
+          listen 80;
+          return 301 https://$$ENV_DISCOURSE_HOSTNAME$request_uri;
+        }
+
+  - file:
+      path: "/etc/nginx/conf.d/outlets/server/http.conf"
+      contents: ""
```
tgxworld, line 16: Instead of setting it to blank, maybe we should remove the file?

templates/web.ssl.template.yml
```diff
@@ -1,56 +1,52 @@
...
+      path: "/etc/nginx/conf.d/outlets/discourse/https.conf"
+      contents: |
+        add_header Strict-Transport-Security 'max-age=31536000';
+
+  - exec:
+      cmd:
+        - |-
+          if [ -f "/proc/net/if_inet6" ] ; then
+            sed -i 's/listen 80;/listen 80;\nlisten [::]:80;/g' /etc/nginx/conf.d/outlets/before-server/redirect-http-to-https.conf
+            sed -i 's/listen 443 ssl;/listen 443 ssl;\nlisten [::]:443 ssl;/g' /etc/nginx/conf.d/outlets/https/https.conf
```
tgxworld, line 51: I think we no longer have `outlets/https` so this should be editing `/etc/nginx/conf.d/outlets/server/https.conf` directly?

templates/web.ssl.template.yml
```diff
@@ -1,56 +1,52 @@
...
+
+  - file:
+      path: "/etc/nginx/conf.d/outlets/server/http.conf"
+      contents: ""
+
+  - file:
+      hook: ssl
+      path: "/etc/nginx/conf.d/outlets/server/https.conf"
+      contents: |
+        listen 443 ssl;
```
tgxworld, line 22:
```suggestion
        listen 443 ssl;
        http2 on;
```

launcher
nbianca: What do you mean? Have them in the base image or have them inline in `*.template.yml`?

templates/web.ssl.template.yml
nbianca: `pups` has no delete command and I found it cleaner to do it like this.

templates/web.ssl.template.yml
nbianca: Do we want to enable http2 on unconditionally?

launcher
tgxworld: Just have them in the base image :)

templates/web.ssl.template.yml
tgxworld: Hmm maybe just `run: rm /etc/nginx/conf.d/outlets/server/http.conf`?

templates/web.ssl.template.yml
tgxworld: I think we already do at the moment so this is ensuring we keep compatibility 

https://github.com/discourse/discourse_docker/blob/20e33fbfd98d3b8d9c57f7a111beff8aa51a5b98/templates/web.ssl.template.yml#L18
https://github.com/discourse/discourse_docker/blob/20e33fbfd98d3b8d9c57f7a111beff8aa51a5b98/templates/web.ssl.template.yml#L24-L25

templates/web.ssl.template.yml
```diff
@@ -1,56 +1,50 @@
...
+          return 301 https://$$ENV_DISCOURSE_HOSTNAME$request_uri;
+        }
+
+  - exec: rm /etc/nginx/conf.d/outlets/server/http.conf
+
+  - file:
+      hook: ssl
+      path: "/etc/nginx/conf.d/outlets/server/https.conf"
+      contents: |
+        listen 443 ssl http2;
```
tgxworld, line 20:
```suggestion
        listen 443 ssl;
        http2 on;
```

Specifying `http2` in the listen directive is deprecated and will emit a warning.

tgxworld (approved):

templates/web.ipv6.template.yml
```diff
@@ -1,8 +1,2 @@
+# This file is deprecated; you can remove it from your app.yml
```
tgxworld, line 1: Let's echo a deprecation message instead as it is highly unlikely that people will see this comment. Let's just set it for removal 1 year later?

tgxworld (approved):

PR #872, opened 2024-09-22
DEV: parameterize config filename

featheredtoast:
Parameterize which filename to write to (write config) and which file is copied from (dockerfile)

### DIFF
--- a/launcher_go/v2/cli_build.go
+++ b/launcher_go/v2/cli_build.go
@@ -36,7 +36,8 @@ func (r *DockerBuildCmd) Run(cli *Cli, ctx *context.Context) error {
 	if err := os.MkdirAll(dir, 0755); err != nil && !os.IsExist(err) {
 		return err
 	}
-	if err := config.WriteYamlConfig(dir); err != nil {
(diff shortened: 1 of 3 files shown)


tgxworld:
@featheredtoast Just from the PR description, I can't deduce why this change is required. Will you be able to update the PR description to reflect the "why"?

featheredtoast:
@tgxworld this one allows for external tools to parameterize where the config.yaml file gets written to, and picked up in a dockerfile, rather than being hardcoded as `config.yaml`

I've got a proof-of-concept write-config-to-docker-compose that could take advantage of this, that allows an export of a 2-container setup to a docker-compose. Without it, I realized there isn't a good way to do something like print multiple config.yaml config files and their dockerfiles to the same location.

Experimental branch here takes advantage of this: https://github.com/featheredtoast/discourse-launcher-extras/commit/33d98b2d86cc15d99784f977a565a95de362a4e7

featheredtoast:
reopened here https://github.com/discourse/launcher/pull/2

PR #874, opened 2024-09-28
DEV: allow for 2 container data services to setup on boot

featheredtoast:
allow postgres templates to assign themselves a DB password allows 2 container setups to be able to create db on boot and setup passwords

move missing folder check to data services when running images without unicorn

standalone unicorn just needs to wait for db to setup via the postgres service start script

### DIFF
--- a/samples/data.yml
+++ b/samples/data.yml
@@ -9,8 +9,10 @@ templates:
 # any extra arguments for Docker?
 # docker_args:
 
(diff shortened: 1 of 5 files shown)


featheredtoast:
closing in favor of #967 first. That proposal does not update any base templates.

I can revisit adding env var for database setup after.

PR #875, opened 2024-09-28
DEV: partially bake in env

featheredtoast:
It may be useful for launcher2 build to bake in some env, allowing for default env to run from a built image.

Introduce defaultBakeEnv as a list of env to bake into images. This allows for folks pulling images to not need to mark every env variable, allowing for some convenience in starting images.


featheredtoast:
@tgxworld this one's an improvement to launcher to bake in instance-agnostic defaults into a build image. They can always be overridden by a cluster at build or run time.

A user's config would only need to define instance-specific environment from a built image, and omit defaults, down to:
```
    environment:
      DISCOURSE_DB_HOST: db
      DISCOURSE_DB_PASSWORD: SOME_SECRET
      DISCOURSE_DEVELOPER_EMAILS: user@example.com
      DISCOURSE_HOSTNAME: localhost
      DISCOURSE_REDIS_HOST: redis
      DISCOURSE_SMTP_ADDRESS: mail
      DISCOURSE_SMTP_PORT: 1025
      DISCOURSE_SMTP_PASSWORD: pa$$word
      DISCOURSE_SMTP_USER_NAME: user@example.com
```

featheredtoast:
Reopened here: https://github.com/discourse/launcher/pull/3

PR #878, opened 2024-10-06
DEV: re-split build vs configure

featheredtoast:
Allow for assets to be precompiled fully in a build image.

The only remaining things for configure is update themes, maxmind download, and css compiling.

Moves ~30 seconds of work from a configure step to a build step, allowing for very quick precompile-on-boot times.

<s>Dev:
* Update web template test from golang to be a symlink.
* Ensure restarting of the unicorn service on migrate/precompile failure</s>

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -234,13 +234,13 @@ run:
       tag: build
       hook: assets_precompile_build
       cmd:
-        - su discourse -c 'bundle exec rake assets:precompile:build'
+        - su discourse -c 'SKIP_DB_AND_REDIS=1 bundle exec rake assets:precompile'
   - exec:
       cd: $home
       tag: precompile
       hook: assets_precompile
       cmd:
-        - su discourse -c 'SKIP_EMBER_CLI_COMPILE=1 bundle exec rake themes:update assets:precompile'
+        - su discourse -c 'bundle exec rake themes:update maxminddb:refresh assets:precompile:css'
 
   - replace:
       tag: precompile


featheredtoast:
@tgxworld this one allows for precompiling of most assets without connecting to any DBs, leading to very little downtime or changes in base `configure` stages.

This means the resulting docker layer is much thinner and those booting up from cold boot (with an image only run with `build`) will also be online much faster.

featheredtoast:
The main piece of this work is in the templates. launcher related patches now here: https://github.com/discourse/launcher/pull/4

templates/web.template.yml
```diff
@@ -62,8 +62,8 @@ run:
           PRECOMPILE_ON_BOOT=1
         fi
         if [ -f /usr/local/bin/create_db ] && [ "$CREATE_DB_ON_BOOT" = "1" ]; then /usr/local/bin/create_db; fi;
-        if [ "$MIGRATE_ON_BOOT" = "1" ]; then su discourse -c 'bundle exec rake db:migrate'; fi
-        if [ "$PRECOMPILE_ON_BOOT" = "1" ]; then SKIP_EMBER_CLI_COMPILE=1 su discourse -c 'bundle exec rake assets:precompile'; fi
+        if [ "$MIGRATE_ON_BOOT" = "1" ]; then su discourse -c 'bundle exec rake db:migrate' || exit 1; fi
+        if [ "$PRECOMPILE_ON_BOOT" = "1" ]; then su discourse -c 'bundle exec rake maxminddb:refresh assets:precompile:css' || exit 1; fi
```
andrewschleifer, line 66: There is no reason for these `if` blocks to be smashed into one line.

templates/web.template.yml
```diff
@@ -62,8 +62,8 @@ run:
           PRECOMPILE_ON_BOOT=1
         fi
         if [ -f /usr/local/bin/create_db ] && [ "$CREATE_DB_ON_BOOT" = "1" ]; then /usr/local/bin/create_db; fi;
```
andrewschleifer, line 64: Should there be `|| exit 1` here, like the below? What happens if `create_db` fails?

andrewschleifer, line 64: Useless trailing `;`

andrewschleifer (dismissed):

andrewschleifer:
Do you have an easy way to test the template output? I am simulating it in my head and ... my head is not reliable.

featheredtoast:
I've narrowed the scope of this, all of the service file updates are now all being done in #967

davidtaylorhq (approved):

PR #879, opened 2024-10-10
Only parse a single line from lscpu to get the Cpu(s) count.

Gullumluvl:
Dear Discourse maintainers,

This is intended to fix an error I was getting with `discourse-setup` on Debian, from line 261:

```stderr
bash: 12
scaling*                   2: syntax error in expression (error token is "scaling*                   2")
```

Caused by the parsing of `lscpu` command retaining *2 lines* instead of just the intended one.

This occurs on my PCs with Debian testing (lscpu from util-linux 2.40.2) and Debian stable (util-linux 2.38.1) because it outputs a line formatted like this (in the "Vendor" section):

    CPU(s) scaling MHz:                   20%

(I also checked on a machine with CentOS, there the format is `CPU MHz:               2700.000`).

To avoid that, I made the regex more specific, but I also added an `exit` statement in `awk`. Probably just one of these fixes could be used.

I believe the script would have worked with the version prior to this commit from a month ago: https://github.com/discourse/discourse_docker/commit/d00c2e75d5884ebefdac3a13f14a7f86b611bda5

Best,

### DIFF
--- a/discourse-setup
+++ b/discourse-setup
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+export LC_ALL=C  # lscpu output is localized.
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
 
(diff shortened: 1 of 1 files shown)


pfaffman:
Maybe this is simpler:

```
lscpu -p|grep -v "#"|wc -l
```

Here's one AI recommended:
```
lscpu | sed -n 's/^CPU(s):\s*//p'
```

These both work inside of a Debian 12 container and Ubuntu 22.04

Gullumluvl:
>     lscpu -p|grep -v "#"|wc -l

This seems best to me, but we must double check with the people using ARM processors, as the commit I mentioned modified this almost exact same line.

Anything expected from me at this point? Sorry I have little experience with contributing to big projects so let me know.

Gullumluvl:
Just rebased my branch on top of main, and amended the commit with a fix for localization.

tgxworld (approved):

tgxworld:
Thank you @Gullumluvl for your contributions.

PR #880, opened 2024-10-15
DEV: slim image, drop unused packages and use cache mounts

featheredtoast:
apt using caches for /var/cache/apt, /var/cache/debconf, and var/lib/apt.

Ensure /usr/share/doc and /usr/share/man do not get saved to the image by mounting temporary folders to paths.

Drop build-essential from installs, in favor of selective cmake, g++, pkg-config, and patch packages.

drop apt-get -y upgrade in dockerfile. We should inherit upgrades from base images. No need to apt-mark hold initscripts now that we're not running `upgrade`

Remove calls to vim as we no longer install vim here. Remove comment for slimming locales as we have now done so.

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -29,70 +29,47 @@ RUN --mount=type=cache,target=/var/cache/debconf,sharing=locked \
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
   --mount=type=cache,target=/var/cache/debconf,sharing=locked \
   --mount=type=cache,target=/var/lib/apt,sharing=locked \
-  --mount=type=tmpfs,target=/usr/share/doc \
-  --mount=type=tmpfs,target=/usr/share/man \
   --mount=type=tmpfs,target=/var/log \
-  apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install gnupg sudo curl fping
-
-RUN sh -c "fping proxy && echo 'Acquire { Retries \"0\"; HTTP { Proxy \"http://proxy:3128\";}; };' > /etc/apt/apt.conf.d/40proxy && apt-get update || true"
-
-RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
-  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
-  --mount=type=cache,target=/var/lib/apt,sharing=locked \
-  --mount=type=tmpfs,target=/usr/share/doc \
-  --mount=type=tmpfs,target=/usr/share/man \
-  --mount=type=tmpfs,target=/var/log \
-  DEBIAN_FRONTEND=noninteractive apt-get install -y locales
-
-ENV LC_ALL en_US.UTF-8
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US.UTF-8
-RUN sed -i "s/^# $LANG/$LANG/" /etc/locale.gen; \
-    locale-gen
-
-RUN install -d /usr/share/postgresql-common/pgdg &&\
-    curl -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc --fail https://www.postgresql.org/media/keys/ACCC4CF8.asc &&\
-    echo "deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] https://apt.postgresql.org/pub/repos/apt ${DEBIAN_RELEASE}-pgdg main" > /etc/apt/sources.list.d/pgdg.list
-
-RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
-  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
-  --mount=type=cache,target=/var/lib/apt,sharing=locked \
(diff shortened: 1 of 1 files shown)


featheredtoast:
@tgxworld this one's the next phase for slimming image size

tgxworld:
How much of the image size are we shaving off here? My first reaction here is that all these changes make the Dockerfile much harder to understand and we really have to be saving a significant amount of disk space for the added complexity to be warranted.

featheredtoast:
This is saving 100MB uncompressed/ 30MB compressed. The majority of the changes here are just adding --mount arguments to the RUN commands.

image/base/Dockerfile
```diff
@@ -40,14 +54,35 @@ RUN install -d /usr/share/postgresql-common/pgdg &&\
...
+  --mount=type=tmpfs,target=/usr/share/doc \
+  --mount=type=tmpfs,target=/usr/share/man \
+  DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends git rsyslog logrotate cron ssh-client less
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
+  --mount=type=cache,target=/var/lib/apt,sharing=locked \
+  --mount=type=tmpfs,target=/usr/share/doc \
+  --mount=type=tmpfs,target=/usr/share/man \
+  DEBIAN_FRONTEND=noninteractive apt-get -y install ca-certificates rsync \
+    cmake g++ pkg-config patch \
```
tgxworld, line 85: I think we can combine this with line 35?

image/base/Dockerfile
```diff
@@ -23,13 +23,27 @@ RUN groupadd --gid 104 postgres &&\
...
-RUN apt-get -y upgrade
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
+  --mount=type=cache,target=/var/lib/apt,sharing=locked \
+  --mount=type=tmpfs,target=/usr/share/doc \
+  --mount=type=tmpfs,target=/usr/share/man \
+  --mount=type=tmpfs,target=/var/log \
+  DEBIAN_FRONTEND=noninteractive apt-get install -y locales
```
tgxworld, line 45: This can be combined with line 35 too.

image/base/Dockerfile
```diff
@@ -63,7 +98,13 @@ RUN sed -i.bak 's/$ModLoad imklog/#$ModLoad imklog/' /etc/rsyslog.conf
...
 RUN dpkg-divert --local --rename --add /sbin/initctl
 RUN sh -c "test -f /sbin/initctl || ln -s /bin/true /sbin/initctl"
-RUN cd / &&\
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
+  --mount=type=cache,target=/var/lib/apt,sharing=locked \
+  --mount=type=tmpfs,target=/usr/share/doc \
+  --mount=type=tmpfs,target=/usr/share/man \
+  --mount=type=tmpfs,target=/root/.npm \
+  cd / &&\
```
tgxworld, line 107: Can we install the apt packages early so we don't have to duplicate these lines here?

image/base/Dockerfile
```diff
@@ -40,14 +54,35 @@ RUN install -d /usr/share/postgresql-common/pgdg &&\
...
+
 # install these without recommends to avoid pulling in e.g.
 # X11 libraries, mailutils
-RUN DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends git rsyslog logrotate cron ssh-client less
-RUN DEBIAN_FRONTEND=noninteractive apt-get -y install autoconf build-essential ca-certificates rsync \
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+  --mount=type=cache,target=/var/cache/debconf,sharing=locked \
+  --mount=type=cache,target=/var/lib/apt,sharing=locked \
+  --mount=type=tmpfs,target=/usr/share/doc \
+  --mount=type=tmpfs,target=/usr/share/man \
```
tgxworld, line 77: Can we try to configure [/etc/dpkg/dpkg.conf.d/01_nodoc](https://unix.stackexchange.com/a/205246) to exclude `/usr/share/doc` and `/usr/share/man`? That will allow us to remove these two lines.

featheredtoast: I looked into this and debian slim's doing this for us already. They do keep the copywright content, but manpages are all clean!

featheredtoast:
it looks like some time ago, a proxy was setup for installing packages, which was causing some complexity in which apt-get install needed to be called multiple times to engage the proxy.

To my knowledge, we no longer use a proxy to install packages, and we're grabbing them directly from upstream, and if so we can safely move and combine apt calls.

featheredtoast:
Re-ordering and merging the apt-get layers here also seems to have squashed the size down impressively to 511.27 MB from 668.47 MB compressed.

tgxworld (approved):

PR #882, opened 2024-11-04
DEV: separate builders

featheredtoast:
slim image by building nginx and imagemagick separately. Copy results to final image.

build jhead from source with no dependencies

remove optipng, and pngcrush as these are disabled workers in discourse image_optim

separate out discourse_web and discourse_release for no database web-only images.

minimize layers by merging RUN commands

prefer `&&\;` over `;\` for failing early on failing commands

Re-orders targets to:
discourse_dependencies (install all common dependencies)
discourse_slim (clone discourse)
discourse_web (bundle install)
discourse_release (install redis+postgres)

Allows for us to target another image as discourse_web to remove postgres+redis from the image. Drops compressed size of slim to 494.22MB.

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -2,34 +2,60 @@
 # VERSION:  release
 
 ARG DEBIAN_RELEASE=bookworm
+FROM discourse/ruby:3.3.4-${DEBIAN_RELEASE}-slim AS builder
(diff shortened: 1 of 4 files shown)


featheredtoast:
will re-open this work in chunks

PR #884, opened 2024-11-07
DEV: slim image by dropping building jhead from source, and dropping optipng

featheredtoast:
jhead from apt pulls in a ton of packages, and optipng isn't used

### DIFF
--- a/image/base/install-oxipng
+++ b/image/base/install-oxipng
@@ -12,7 +12,11 @@ case "${dpkgArch##*-}" in
 esac
 
 # Install other deps
-apt -y -q install advancecomp jhead jpegoptim libjpeg-turbo-progs optipng
+apt -y -q install advancecomp jpegoptim libjpeg-turbo-progs
+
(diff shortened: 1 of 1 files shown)


tgxworld:
How much do we save by compiling jhead ourselves?

featheredtoast:
This change is 8MB compressed. It'd be very nice to have a github action be able to report this for us at some point.

featheredtoast:
main: 648.38MB
compiling jhead ourselves: 640.2MB
compiling jhead ourselves + dropping optipng: 640.14MB

tgxworld (approved):

image/base/install-oxipng
```diff
@@ -12,7 +12,11 @@ case "${dpkgArch##*-}" in
 esac
 
 # Install other deps
-apt -y -q install advancecomp jhead jpegoptim libjpeg-turbo-progs optipng
+apt -y -q install advancecomp jpegoptim libjpeg-turbo-progs
```
tgxworld, line 15: Not really related to this PR but I did wonder why we are installing various jpeg packages in the `install-oxipng` script which I would think should only deal with `oxipng`.

PR #886, opened 2024-11-08
upload launcher2 to github release

featheredtoast:
create release on tag push of type `v*.*.*`

uploads both tar.gz and md5.

on tag creation EG, v2.0.0 this github action will:

* create a release of the same name -- v2.0.0
* Upload gzipped binary assets for each OS/arch, along with their md5 hashes

This does not use goreleaser, or go-release-action. Creation is simple with setup go, and compilation in the shell. It then uses github cli to upload the release assets.

### DIFF
--- a/.github/workflows/launcher_go.yml
+++ b/.github/workflows/launcher_go.yml
@@ -43,3 +43,37 @@ jobs:
       - name: Run tests
         working-directory: ./launcher_go/v2
         run: go test ./...
+
+  build_release:
+    if: github.ref == 'refs/heads/main'
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    needs: [lint, test]
+    strategy:
+      matrix:
+        goos: [linux, darwin]
+        goarch: ["386", amd64, arm64]
+        exclude:
+          - goarch: "386"
+            goos: darwin
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/launcher_go.yml
```diff
@@ -2,6 +2,8 @@ name: launcher_go
 
 on:
   push:
+    tags:
+      - "v*.*.*"
```
tgxworld, line 6: I don't think we should be using tags here to push out a release. This repository contains more than just `launcher` so a git tag would make little sense for the `discourse_docker` related stuff. Are we able to use some sort of version file here so that a new release is created when the version in the file is bumped?

featheredtoast: sure let me update to watch for that 👍

featheredtoast: So, short answer no it's not going to work exactly as you requested. But I've updated it so we push assets to a single release.

Github releases require a git tag, and if we don't want to provide one, we need to create a dummy tag for the release to pin to, and create a single `launcher2` release. We can then overwrite assets within that release, and that single release becomes our go-to 'latest' release for the assets.

featheredtoast:
Updated so we're using GH CLI to upload release assets to a single static release instead of creating and uploading on a tag creation.

featheredtoast:
closing in favor of https://github.com/discourse/launcher/pull/1

PR #887, opened 2024-11-08
FEATURE: print out compressed image sizes for amd64

featheredtoast:
Print out compressed image sizes for amd64 images, in PRs, adds a step that prints a line such as:

`current slim: 676.45 release: 1007.99. new slim: 676.46 release: 1008.04`

making it easier to compare image sizes.

remove jobs from pushing on forks.

Remove jobs from scheduled jobs from running on forks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ env:
 
 jobs:
   timestamp:
-    if: (github.event_name != 'schedule' || github.event.repo.name == 'discourse/discourse_docker')
+    if: (github.event_name != 'schedule' || github.repository == 'discourse/discourse_docker')
     runs-on: ubuntu-latest
     outputs:
       timestamp: ${{ steps.timestamp.outputs.timestamp }}
@@ -34,7 +34,7 @@ jobs:
           echo "timestamp=$timestamp" >> $GITHUB_OUTPUT
 
   base:
-    if: (github.event_name != 'schedule' || github.event.repo.name == 'discourse/discourse_docker')
+    if: (github.event_name != 'schedule' || github.repository == 'discourse/discourse_docker')
     # `unbuntu-22.04-8core` for arch amd64 non-scheduled builds
     # `unbuntu-22.04` for arch amd64 scheduled builds
     # `unbuntu-22.04-8core-arm` for arch arm64 non-scheduled builds
@@ -123,7 +123,7 @@ jobs:
           echo "current slim: ${CURRENT_SLIM} release: ${CURRENT_RELEASE}. new slim: ${NEW_SLIM} release: ${NEW_RELEASE}"
 
       - name: push to dockerhub
-        if: github.ref == 'refs/heads/main' && github.event.repo.name == 'discourse/discourse_docker'
+        if: github.ref == 'refs/heads/main' && github.repository == 'discourse/discourse_docker'
         env:
           DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
         run: |
@@ -134,7 +134,7 @@ jobs:
           docker push discourse/discourse_dev:${{ env.TIMESTAMP }}-${{ matrix.arch }}
 
       - name: Push discourse/base:aarch64 image for backwards compatibility
-        if: (github.ref == 'refs/heads/main') && (matrix.arch == 'arm64') && (github.event.repo.name == 'discourse/discourse_docker')
+        if: (github.ref == 'refs/heads/main') && (matrix.arch == 'arm64') && (github.repository == 'discourse/discourse_docker')
         run: |
           docker tag discourse/base:2.0.${{ env.TIMESTAMP }}-main-${{ matrix.arch }} discourse/base:aarch64
           docker push discourse/base:aarch64
@@ -144,7 +144,7 @@ jobs:
     needs: [base, timestamp]
     env:
       TIMESTAMP: ${{ needs.timestamp.outputs.timestamp }}
-    if: github.ref == 'refs/heads/main' && github.event.repo.name == 'discourse/discourse_docker'
+    if: github.ref == 'refs/heads/main' && github.repository == 'discourse/discourse_docker'
     steps:
       - name: create and push multiarch manifests
         run: |
@@ -231,7 +231,7 @@ jobs:
         run: |
           docker images discourse/discourse_test
       - name: push to dockerhub
-        if: success() && (github.ref == 'refs/heads/main') && (github.event.repo.name == 'discourse/discourse_docker')
+        if: success() && (github.ref == 'refs/heads/main') && (github.repository == 'discourse/discourse_docker')
         env:
           DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
         run: |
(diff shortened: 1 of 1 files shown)


featheredtoast:
Screenshot 2024-11-08 at 11 51 26 AM

This PR's print

.github/workflows/build.yml
```diff
@@ -33,6 +34,7 @@ jobs:
           echo "timestamp=$timestamp" >> $GITHUB_OUTPUT
 
   base:
+    if: (github.event_name != 'schedule' || github.repository == 'discourse/discourse_docker')
```
tgxworld, line 37: I'm not sure I follow here. Why are we skipping building of the base image when it is a scheduled job?

.github/workflows/build.yml
```diff
@@ -103,8 +105,25 @@ jobs:
...
+          CURRENT_RELEASE=$(docker manifest inspect -v discourse/base:release | jq -r '.[0].SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_SLIM=$(docker manifest inspect -v --insecure localhost:5002/base:2.0.${{ env.TIMESTAMP }}-slim-${{ matrix.arch }} |  jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_RELEASE=$(docker manifest inspect -v --insecure localhost:5002/base:2.0.${{ env.TIMESTAMP }}-main-${{ matrix.arch }} | jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          echo "current slim: ${CURRENT_SLIM} release: ${CURRENT_RELEASE}. new slim: ${NEW_SLIM} release: ${NEW_RELEASE}"
+
       - name: push to dockerhub
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && github.repository == 'discourse/discourse_docker'
```
tgxworld, line 126: I'm not sure if I agree here. I think forks should just disable actions if they don't want to run them rather than the main repository having to include this conditional.

.github/workflows/build.yml
```diff
@@ -103,8 +105,25 @@ jobs:
         run: |
           docker images discourse/base
 
+      - name: Print compressed summary
+        if: github.event_name == 'pull_request' && matrix.arch == 'amd64'
+        run: |
+          # Push to local repo to compare sizes
+          docker run --quiet -d --rm -p 5002:5000 registry:2
+          sleep 5
```
tgxworld, line 113: Instead of doing this, I think we can follow https://docs.docker.com/build/ci/github-actions/local-registry/ instead.

.github/workflows/build.yml
```diff
@@ -103,8 +105,25 @@ jobs:
...
+          # multiarch manifest is an array of schemas - [0] is amd64, [1] is arch64: Compare amd64.
+          CURRENT_SLIM=$(docker manifest inspect -v discourse/base:slim | jq -r '.[0].SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          CURRENT_RELEASE=$(docker manifest inspect -v discourse/base:release | jq -r '.[0].SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_SLIM=$(docker manifest inspect -v --insecure localhost:5002/base:2.0.${{ env.TIMESTAMP }}-slim-${{ matrix.arch }} |  jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_RELEASE=$(docker manifest inspect -v --insecure localhost:5002/base:2.0.${{ env.TIMESTAMP }}-main-${{ matrix.arch }} | jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          echo "current slim: ${CURRENT_SLIM} release: ${CURRENT_RELEASE}. new slim: ${NEW_SLIM} release: ${NEW_RELEASE}"
```
tgxworld, line 123: I think including the units will be better here.

.github/workflows/build.yml
featheredtoast: I'm of the opinion it's better for a fork to be able to get a passing CI run for experimenting. Forks may be interested in other jobs that are not push-to-`discourse/base`, such as test builds, running tests, and linting and we should consider allowing that.

We can remove it, but it was quite useful here for this work, so I thought to propose it.

.github/workflows/build.yml
featheredtoast: skipping when it's a scheduled job and is a fork - this prevents a bunch of jobs on new forks until actions are disabled but will remove for now from the other comment.

.github/workflows/build.yml
featheredtoast: Oh didn't know services were available to us, very cool!

.github/workflows/build.yml
```diff
@@ -103,6 +108,21 @@ jobs:
...
+          CURRENT_SLIM=$(docker manifest inspect -v discourse/base:slim | jq -r '.[0].SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          CURRENT_RELEASE=$(docker manifest inspect -v discourse/base:release | jq -r '.[0].SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_SLIM=$(docker manifest inspect -v --insecure localhost:5000/base:2.0.${{ env.TIMESTAMP }}-slim-${{ matrix.arch }} |  jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          NEW_RELEASE=$(docker manifest inspect -v --insecure localhost:5000/base:2.0.${{ env.TIMESTAMP }}-main-${{ matrix.arch }} | jq -r '.SchemaV2Manifest.layers[] | .size / 1024 / 1024 | .*100 | round/100' | awk '{print $0; sum+= $0}; END {print sum}' | tail -n 1)
+          echo "current slim: ${CURRENT_SLIM}MB release: ${CURRENT_RELEASE}MB. new slim: ${NEW_SLIM}MB release: ${NEW_RELEASE}MB"
```
tgxworld, line 124: Small non-blocking comment. If we are ever trying to split up the build, alot of the commands here can be executed in parallel.

tgxworld (approved):

.github/workflows/build.yml
featheredtoast: took a quick glance and it seems like steps can't be parallelized, so we'd have to split it up further into tasks but it'd be nice to do certainly 👍

PR #891, opened 2024-11-19
Update launcher script to download and call launcher rewrite

featheredtoast:
download, checksum and use https://github.com/discourse/launcher

support update arg to re-download launcher binary.

keep most of the prereqs function in place. Use hello-world image to check docker install.

check for autocompletion and squelch output if we're calling autocomplete.

Remove checks on config existence and config permissions. If we need this check it should go into the launcher app.

update overlay check to include overlayfs which is what comes up on default latest docker on desktop versions

### DIFF
--- a/launcher
+++ b/launcher
@@ -180,14 +180,14 @@ update() {
 }
 
 case "$1" in
-	  update)
-        echo "updating launcher..." >&2
-		    update
-		    exit 0
-		    ;;
-	  *)
-		    test -f "${BINDIR}/launcher" || update
-        echo "run ./launcher update to update launcher" >&2
-        check_prereqs
-		    exec "${BINDIR}/launcher" "$@"
+  update)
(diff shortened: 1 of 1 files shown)


launcher
```diff
@@ -175,721 +67,127 @@ compare_version() {
...
+        echo "updating launcher..." >&2
+		    update
+		    exit 0
+		    ;;
+	  *)
+		    test -f "${BINDIR}/launcher" || update
+        echo "run ./launcher update to update launcher" >&2
+        check_prereqs
+		    exec "${BINDIR}/launcher" "$@"
 esac
```
andrewschleifer, line 193: This `case` statement uses mixed tabs and spaces for indentation.

andrewschleifer (approved):
Whitespace issues, otherwise LGTM.

PR #894, opened 2024-12-16
DEV: Bump rust version to 1.82.0

tgxworld:
This is to resolve an error we are seeing in our arm64 dev builds

```
24 49.46 error: package `apple-xar v0.20.0` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.75.0
24 49.46 Either upgrade to rustc 1.81 or newer, or use
```


CvX (approved):

tgxworld:
Thank you @CvX 👍

PR #895, opened 2024-12-17
Remove unnecessary base dependencies that bloat layer size

tgxworld:
This change reduces the layer for this `RUN` command from 1GB to 729MB
as shown in the `docker history <image name>` output.

Packages removed are:

`pkg-config`, `libxslt-dev` and `libxml2-dev`: This listed in the [nokogiri's gem docs](https://nokogiri.org/tutorials/installing_nokogiri.html#installing-using-the-packaged-libraries) but we rely on Nokogiri's packaged version. 

`libtool`, `parallel`: I found no good reason for these two packages to exist but they add significant bloat.

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -25,6 +25,14 @@ RUN gpg --import /tmp/nginx_public_keys.key &&\
     rm /tmp/nginx_public_keys.key &&\
     /tmp/install-nginx
 
+FROM builder AS thpfoff_builder
+
+# This tool allows us to disable huge page support for our current process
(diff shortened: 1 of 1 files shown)


featheredtoast:
Some of the gems we use compile at install time which means we're currently dependent on compulers unfortunately

tgxworld:
> Some of the gems we use compile at install time which means we're currently dependent on compulers unfortunately

I've got a fix incoming :)

tgxworld:
OK this turned out to be more difficult that I have imagined. Going to just drop this for now.

PR #896, opened 2024-12-17
Remove updating of rubygems

tgxworld:
This adds an additional layer of 30mb for no reason. We don't need to be
running the latest version of rubygems all the time.

```
tgxworld ~/work/discourse_docker [skip_updating_ruby_gems] $ docker history --no-trunc discourse/base:release | grep 'gem update --system'
<missing>                                                                 5 hours ago    RUN |1 DEBIAN_RELEASE=bookworm /bin/sh -c echo 'gem: --no-document' >> /usr/local/etc/gemrc &&    gem update --system # buildkit                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                29.8MB    buildkit.dockerfile.v0
```

Even if we need to update rubygems, it should be updated in https://github.com/discourse/docker-ruby

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -127,9 +127,6 @@ RUN /tmp/install-redis
 ADD install-oxipng /tmp/install-oxipng
 RUN /tmp/install-oxipng
 
-RUN echo 'gem: --no-document' >> /usr/local/etc/gemrc &&\
-    gem update --system
-
 RUN gem install pups --force &&\
     mkdir -p /pups/bin/ &&\
     ln -s /usr/local/bin/pups /pups/bin/pups


image/base/Dockerfile
```diff
@@ -127,9 +127,6 @@ RUN /tmp/install-redis
 ADD install-oxipng /tmp/install-oxipng
 RUN /tmp/install-oxipng
 
-RUN echo 'gem: --no-document' >> /usr/local/etc/gemrc &&\
```
tgxworld, line 130: Note that the ruby image that this image is based off already does this: https://github.com/discourse/docker-ruby/blob/832c4cbc71ae9ad37b7206d83b73e473472a2e30/Dockerfile#L31-L34

tyb-talks (approved):

PR #898, opened 2024-12-18
Print docker history summary in build actions

tgxworld:
It is useful to know what the size of each layer is after the image has
been built.


pento (approved):

tgxworld:
Thank you @pento 👍

PR #908, opened 2025-01-08
Separate nginx build step

featheredtoast:
redo #897 

Adds nginx-common, which adds boilerplate files such as `/etc/nginx` directory


featheredtoast:
@tgxworld the previous iteration looked good, just needed nginx-common to also be installed in the base image.

tgxworld (approved):

PR #909, opened 2025-01-10
FEATURE: Update base image and set default to postgresql 15

mwaniki-wairungu:
This updates the default PostgreSQL version to 15.
  * image/base: update default postgres version to 15
  * launcher: bump base image
  * postgres.template.yml: update default version to 15
  * postgres.15.template.yml: current version template

Legacy version templates:
  * postgres.13.template.yml
  * postgres.12.template.yml
  * postgres.10.template.yml
  * postgres.9.5.template.yml

### DIFF
--- a/launcher
+++ b/launcher
@@ -92,7 +92,7 @@ kernel_min_version='4.4.0'
 config_file=containers/"$config".yml
 cidbootstrap=cids/"$config"_bootstrap.cid
 local_discourse=local_discourse
-image="discourse/base:2.0.20250114-0014"
+image="discourse/base:2.0.20250127-0958-pg-15"
 docker_path=`which docker.io 2> /dev/null || which docker`
 git_path=`which git`
 
--- a/templates/postgres.template.yml
+++ b/templates/postgres.template.yml
@@ -28,7 +28,7 @@ run:
      contents: |
         #!/bin/sh
         exec 2>&1
-        HOME=/var/lib/postgresql USER=postgres exec thpoff chpst -u postgres:postgres:ssl-cert -U postgres:postgres:ssl-cert /usr/lib/postgresql/13/bin/postmaster -D /etc/postgresql/13/main
+        HOME=/var/lib/postgresql USER=postgres exec thpoff chpst -u postgres:postgres:ssl-cert -U postgres:postgres:ssl-cert /usr/lib/postgresql/15/bin/postmaster -D /etc/postgresql/15/main
 
   - file:
      path: /etc/service/postgres/log/run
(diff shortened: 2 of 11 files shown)


tgxworld:
@mwaniki-wairungu The test failure looks legit here.

templates/postgres.template.yml
```diff
@@ -131,82 +136,90 @@ run:
...
+  - replace:
+      filename: "/etc/postgresql/15/main/pg_hba.conf"
+      from: /^host.*replication.*all.*127.*$/
+      to: "host replication all 0.0.0.0/0 md5"
+
+  # allow replication to connect in with md5/scram auth (IPv6)
+  - replace:
+      filename: "/etc/postgresql/15/main/pg_hba.conf"
+      from: /^host.*replication.*all.*::1\/128.*$/
+      to: "host replication all ::/0 md5"
```
tgxworld, line 221: We don't support PG replication in the self hosted set up so I'm wondering if we can just drop these lines.

mwaniki-wairungu: I've removed them. Thanks for spotting that.

mwaniki-wairungu:
> The test failure looks legit here.

Fixed. The tests are now green.

tgxworld (approved):

tgxworld:
@mwaniki-wairungu Just confirming but did we test on a self-hosted installation that we are able to upgrade from PG 13 to PG 15?

mwaniki-wairungu:
> did we test on a self-hosted installation that we are able to upgrade from PG 13 to PG 15?

Yes. The upgrade was successfully tested on a self-hosted site populated with some test data.

tgxworld:
> Yes. The upgrade was successfully tested on a self-hosted site populated with some test data.

Thank you! The changes look good to me 👍

launcher
```diff
@@ -92,6 +92,7 @@ kernel_min_version='4.4.0'
 config_file=containers/"$config".yml
 cidbootstrap=cids/"$config"_bootstrap.cid
 local_discourse=local_discourse
+# TODO: replace base image before merging
```
tgxworld, line 95: @mwaniki-wairungu I just realised a problem you have here. The new base image with PG 15 installed is only built and made available roughly 30 minutes after this commit lands on the `main` branch. However it is possible for self hosters to be using the new templates before we are able to update the tag in `launcher`. With that in mind, I think we need some sort of conditionals in the existing template where we account for the fact that either PG 15 or PG 13 packages may be present in the base image?

mwaniki-wairungu: During the [PG 13 update](https://github.com/discourse/discourse_docker/pull/493/files), we built a temporary 'placeholder' image from the `pg13` branch which was then referenced in launcher. After the changes were merged and the new 'official' image was built, a [follow-up commit](https://github.com/discourse/discourse_docker/pull/494/files) updated the ref accordingly.

I think we should use a similar approach.

mwaniki-wairungu: Thanks @tgxworld for [building the temporary image](https://github.com/discourse/discourse_docker/commit/ab957b2f64b8252e18a657e2e5abe0cff087b38a).

I'll update the refs and merge early next week.

PR #918, opened 2025-01-17
Add new Nginx config to base image

nbianca:
This commit moves the Nginx config from discourse/discourse repository
and updates it with several 'include' statements that serve as extension
outlets.

This change was made to place the config closer to where it is used and
implements an alternative to the "find & replace" hacks.

It is unused at this moment, but all templates and samples will be
updated in a future commit.


tgxworld (approved):

tgxworld:
@nbianca Do confirm that the existing web template will override the nginx conf file that has been added to the base image: https://github.com/discourse/discourse_docker/blob/136c63890674b95df1327d24270c55e4ef8e87a8/templates/web.template.yml#L135

nbianca:
> Do confirm that the existing web template will override the nginx conf file that has been added to the base image:

`cp` will overwrite the target file without warning if that happens, so it should work as before.

PR #920, opened 2025-01-21
DEV: switch amd64 builds to debian-12 runners

fitzy101:
We can use our self-hosted action/runner machines to build the amd64 Docker images, falling back to the Github action/runner machines for the arm64 builds.


fitzy101:
That expression is _ghastly_.

tgxworld (approved):

PR #924, opened 2025-01-23
DEV: bump timeout for non-scheduled builds

fitzy101:
The most recent build timed out at 30 minutes. Bump timeout temporarily to get a gauge on how long self-hosted builds take (push to Dockerhub looks to be slower than Github-hosted runners).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       matrix:
         arch: [amd64, arm64]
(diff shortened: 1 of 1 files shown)


fitzy101:
See https://github.com/discourse/discourse_docker/actions/runs/12939329448/job/36093269021.

fitzy101:
The successful `amd64` build here took 30min 2s :face_exhaling: 

https://github.com/discourse/discourse_docker/actions/runs/12940557907/job/36095025663

So it looks like we _just_ exceeded the 30min timeout

tgxworld (approved):

PR #925, opened 2025-01-26
Fix TLS in socketed template

nta:
After 7d548ad4ae3022863262f485505a32df13df1983, the replacement pattern in web.socketed.template.yml no longer matches, making nginx.https.sock not exist.

As such, remove the `http2` field from the `listen` directive to match aforementioned commit.


tgxworld (approved):

tgxworld:
@nta Thank you for fixing this 👍

PR #938, opened 2025-02-04
FIX: web.ssl.template compat with latest core

davidtaylorhq:
https://github.com/discourse/discourse/commit/649505d869ff7a04d894fb2322428d0cfb82100e added some new content between `listen 80;` and `gzip on;`, which broke these replacements.

This commit reorders & updates the replacements so that they should work reliably with versions of core before and after that change.

- Move the `/server.+{/` replacement to the end, so that it doesn't introduce a duplicate `listen 80;` which will get in the way of the other replacements
- Update the `listen 80; gzip on;` replacement to just look for `listen 80;`
- Update the `listen 80;\s+listen \[::\]:80; gzip onl` replacement to just `listen 80;\s+listen \[::\]:80;`
- Swap those two around, so that only one of them can ever run

### DIFF
--- a/templates/web.ssl.template.yml
+++ b/templates/web.ssl.template.yml
@@ -4,26 +4,17 @@ run:
        - "mkdir -p /shared/ssl/"
   - replace:
      filename: "/etc/nginx/conf.d/discourse.conf"
-     from: /server.+{/
-     to: |
-       server {
-         listen 80;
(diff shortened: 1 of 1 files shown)


nbianca (approved):

Firefishy:
The final server block `listen 80;` used the redirect isn't enabled for IPv6 regardless of enabling `web.ssl.template.yml`
I'm don't see issue just unconditionally turning on IPv6 listening.

PR #939, opened 2025-02-11
Update firefox tar format

davidtaylorhq:
Firefox started distributing `.tar.xz` instead of `.tar.bz2`


CvX (commented):
CI says 🙅 😅 

```
26.97 bzip2: (stdin) is not a bzip2 file.
26.97 tar: Child returned status 2
26.97 tar: Error is not recoverable: exiting now
```

Flink:
We need to use `-J` instead of `-j` for the tar commands :sweat_smile:

CvX (approved):

PR #940, opened 2025-02-17
Retry assets:precompile up to three times

nbianca:
`assets:precompile` may fail because of `Zlib::BufError`. This might be a sprockets/Zlib bug.

See https://github.com/ruby/zlib/issues/49


nbianca:
Superseded by https://github.com/discourse/discourse/pull/31398.

PR #949, opened 2025-03-10
DEV: Update standalone.yml

phillcoxon:
Improve SMTP guidance in the config file

Updated the template advising that the sending domain or sub-domain and DISCOURSE_NOTIFICATION_EMAIL address should be authenticated with SMTP providers to ensure email delivery. 

Also provided a direct link to info on the `discourse-doctor` command that can assist with testing and debugging email configurations.

### DIFF
--- a/samples/standalone.yml
+++ b/samples/standalone.yml
@@ -61,6 +61,15 @@ env:
   DISCOURSE_DEVELOPER_EMAILS: 'me@example.com,you@example.com'
 
   ## TODO: The SMTP mail server used to validate new accounts and send notifications
+  #
+  # Most SMTP providers will require an authenticated domain or authenticated email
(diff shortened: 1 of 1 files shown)


phillcoxon:
Hi!

Please let me know if you'd me to make any changes to this PR.  

When accepted I can do a similar PR for the web_only.yml config template. 

As I'm not familiar with the codebase please also let me know if there are other templates that should updating?  For example I see standalone.yml and web_only.yml config files in https://github.com/discourse/launcher/blob/main/v2/test/containers

Related post on Meta: https://meta.discourse.org/t/troubleshoot-email-on-a-new-discourse-install/16326/554

@SouthpawKB  for awareness

awesomerobot (approved):

PR #960, opened 2025-05-09
DEV: Set fixed playwright browsers path for test Docker image

tgxworld:


CvX (approved):

davidtaylorhq:
I think we should revert this in favour of https://github.com/discourse/discourse_docker/pull/956

PR #963, opened 2025-05-12
DEV: include base image in web template

featheredtoast:
This will move base image bumps from launcher into web.template.yml


tgxworld (approved):
Change looks safe to me but do keep an eye out on meta in case any potential breaking change.

PR #966, opened 2025-05-18
FEATURE: introduce web_only images

featheredtoast:
web_only images would not install postgres or redis, minimizing image sizes for setups that do not require those components.

release and dev images still install postgres and redis, but do it later in the process.


image/base/Dockerfile
```diff
@@ -145,14 +143,6 @@ RUN gem install pups --force &&\
 ADD thpoff.c /src/thpoff.c
 RUN gcc -o /usr/local/sbin/thpoff /src/thpoff.c && rm /src/thpoff.c
 
-# clean up for docker squash
```
featheredtoast, line 148: We no longer use docker squash - that has been gone for a long time. Here, we now desire /tmp to stick around as a way of keeping /tmp/install-redis around for installing later within a build for discourse-dev.

image/base/Dockerfile
```diff
@@ -130,7 +129,6 @@ ADD install-jemalloc /tmp/install-jemalloc
 RUN /tmp/install-jemalloc
 
 ADD install-redis /tmp/install-redis
```
nbianca, line 131: This can be moved too.

image/base/Dockerfile
nbianca: I wonder if we could delete these after the `RUN` command that generates them to keep the image small.

image/base/Dockerfile
featheredtoast: We'd need a way to add it to the dev image as well when we build that, as that also needs to be able to run install-redis in a different context.

My concern with copying is that any changes to install-redis could get out of sync in an update.

image/base/Dockerfile
featheredtoast: The other option we have is to add a tmpfs have a for these during the build.

Note that I'm keeping /tmp around to ensure install-redis remains as needed - it's called in both this image as well as the dev image.

featheredtoast:
@nbianca any update here?

andrewschleifer (approved):

PR #970, opened 2025-06-14
Replace launcher with a Python script

DrChat:
This change completely replaces the `launcher` bash script with an equivalent written in Python.
This Python script is intended to have exactly identical functionality to the previous script (and the CLI should be 100% backwards compatible).

I made this change primarily to create a script that can be ran natively on Windows and to make some optimizations to the startup process.

The following features were added:
* Template YAML files can now have a `build` block that contains commands that will be ran when the container is built.


featheredtoast:
Hey thanks for the suggestion -- at a glance this is not something we can accept unfortunately.

There have been some internal discussions about ways to get done what you're looking to do here, and it's something we've had our eye on. TLDR, it's a bit more complicated than just making things happen at container boot.

There are a few assumptions you're making here that would concern me:
* One added assumption you're adding in your proposal is that python3 is available to a system building Discourse, which might cause some friction with users on systems without it. The purpose of the bash script initially is to have very few assumptions other than a linux-y system. I understand you'd like to support windows, but trading the dependency for a python3 dependency isn't the way forward.
* Moving all the build code into running on boot would initialize the container twice -- once on build, and then once again on boot.
* The pups init scripts make assumptions that they can connect to a running database, which complicates a docker build that doesn't have access to mount directories.
* One of the reasons for launcher is to get a prebuild step out of the way so the resulting container's boot is quick. Forcing a full compile at boot would be very slow, and for containers that need to be brought up quickly (EG, if you are using Discourse in a cloud autoscaling environment) it'd bring agility down a ton.
* No control over migrations, full migrations run every time an app container starts.

If you'd like to take a peek at what we have brewing, [the current proposal is a rewrite using go](https://github.com/discourse/launcher). That strives to take into account the above and is a current work in progress.

DrChat:
Okay fair enough. This script was fairly hacky :)

The main thing I was aiming for was to eliminate the requirement to use the `launcher` script to start Discourse.
I'm deploying to an Azure application service where I only have the option to start the service from an image (without a wrapper script).
At first glance, it _looks_ like launcher V2 can support this, which is great!

I do want to note (and maybe you have already incorporated this into launcher V2) that a lot of the startup steps can be done at build-time.
For example, I have moved the following steps to container build (and this is the current process for my production deployments):
* Cloning/pulling the Discourse repository
* Installing yarn/pnpm dependencies
* Pulling plugins
* Running `assets:precompile:build`

This took container startup times from ~5-10 minutes to ~1-2 minutes, and the most expensive parts of the build process are done on my development box.

One question: how far along is launcher V2? Is it ready to use?

featheredtoast:
I'd say V2 is *functional*, but it's very much not officially supported: Very much a **here be dragons** kind of disclaimer if you decide to start down that path.

I'm also not targeting windows in this proposal yet.

`launcher build app.yml`. push the resulting image to somewhere you're pulling in Azure.

> I do want to note (and maybe you have already incorporated this into launcher V2) that a lot of the startup steps can be done at build-time.

Indeed, that is how it's working under the hood. It's splitting up build-steps from need-a-database steps, exactly.

---

Your steps would then be `launcher build app.yml`, and push the resulting image somewhere. Unlike `launcher bootstrap` you don't need access to the DB -- this can be built anywhere. You'll need to also set the following 3 env vars to tell containers initialize at boot:
```
MIGRATE_ON_BOOT=1
PRECOMPILE_ON_BOOT=1
CREATE_DB_ON_BOOT=1
```

PR #973, opened 2025-07-07
DEV: always change postgres permissions on existing data directory

featheredtoast:
If permissions of the initial data directory get changed, the install script is unable to reset the permissions to the expected user.

The intent of the chown command is to ensure the permissions of the database and run directory are consistent and known. This allows it to do that on an existing data directory where this is more likely to occur.

This uses a more consistent find command by moving chown command under exec Do not rely on a file glob which can error on too many files.


pfaffman:
This is great. I've had to do this several times myself on upgrades from a very old container and have helped several people do it on meta.

Thanks!

tgxworld (approved):

PR #974, opened 2025-07-09
Provide instructions for removing bundled plugin clone config

davidtaylorhq:
We will soon start bundling many more plugins in Discourse core, which means that existing `git clone` lines in container YAMLs will need to be removed. This commit adds a more informative message for this situation.

### DIFF
--- a/launcher
+++ b/launcher
@@ -23,6 +23,10 @@ usage () {
   exit 1
 }
 
+BUNDLED_PLUGINS=(
+  "discourse-reactions"
+)
+
 # for potential re-exec later
 SAVED_ARGV=("$@")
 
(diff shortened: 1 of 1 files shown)


pfaffman:
Nice work! Better still would be to just have `sed` remove those automatically and have them run ./launcher again. 

Something like this . . . (maybe commenting them out would be better?)

```
clean_bundled_plugins_from_config() {
  config_data=$(cat "$config_file")
  for plugin in "${BUNDLED_PLUGINS[@]}"; do
    if grep -q "git clone https://github.com/discourse/$plugin" "$config_file"; then
      echo "---"
      echo "NOTICE: The plugin '$plugin' is now bundled with Discourse and does not need to be in your container config."
      echo "Removing 'git clone https://github.com/discourse/$plugin' from $config_file"
      echo "For more information, see https://meta.discourse.org/t/373574"
      echo "---"
      sed -i.bak "/git clone https:\/\/github.com\/discourse\/$plugin/d" "$config_file"
    fi
  done
}
```

CvX (approved):

PR #989, opened 2025-07-30
FIX: improve postgres upgrade reliability

jcharaoui:
On some systems the /shared/postgres_data directory might be a mount point, so in order to avoid failure in this scenario, move the postgres data files directly instead of manipulating the data directory.

see https://meta.discourse.org/t/infinite-loop-during-postgres-upgrade/91767/8?u=lavamind

### DIFF
--- a/templates/postgres.15.template.yml
+++ b/templates/postgres.15.template.yml
@@ -132,8 +132,10 @@ run:
            exit 1
          fi
 
-         mv /shared/postgres_data /shared/postgres_data_old
-         mv /shared/postgres_data_new /shared/postgres_data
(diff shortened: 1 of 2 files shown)


tgxworld:
@Supermathie @fitzy101 This looks OK to me but can you also review it please? Thank you!

fitzy101:
> @Supermathie @fitzy101 This looks OK to me but can you also review it please before I merge? Thank you!

Looks OK to me too.

fitzy101 (approved):

tgxworld:
Thank you @jcharaoui 👍

PR #992, opened 2025-08-02
letsencrypt updates: renew location for .well-known, add support for multiple hostnames

featheredtoast:
add comma-separated DISCOURSE_HOSTNAME_ALIASES to handle multiple aliases
for letsencrypt domain generation over env vars

FIX: add letsencrypt renew location for .well-known and allow for multi-domain renewal
Do not redirect to https on .well-known location for any domain
so letsencrypt cert renewals can work properly.

FIX: better listening for ipv6 for ipv4 only
Current letsencrypt assumes ipv6 in config. Check for ipv6 before listening

### DIFF
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -40,7 +40,6 @@ run:
 
         events {
           worker_connections 768;
-          # multi_accept on;
         }
 
         http {
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -69,7 +69,6 @@ run:
           sed -i 's/listen 80;/listen 80;\nlisten [::]:80;/g' /etc/nginx/letsencrypt.conf
         fi
 
-
         sed -Ei "s/^#?ACCOUNT_EMAIL=.+/ACCOUNT_EMAIL=${LETSENCRYPT_ACCOUNT_EMAIL}/" \
           /shared/letsencrypt/account.conf
 
--- a/templates/web.letsencrypt.ssl.template.yml
+++ b/templates/web.letsencrypt.ssl.template.yml
@@ -90,10 +90,10 @@ run:
         /usr/sbin/nginx -c /etc/nginx/letsencrypt.conf
 
         extra_domains() {
-        if [ -n "$DISCOURSE_HOSTNAME_ALIASES" ]; then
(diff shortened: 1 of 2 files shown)


pfaffman:
Will this let all requests get through and not just those for the discourse hostname? I think so, but I'm not quite sure

I'm trying to get certs for additional domains. The old way no longer works.

pfaffman:
DUDE! That is so awesome! I can't wait to check it out.

andrewschleifer (commented):
Can we get rid of this line? @tgxworld git blame tracks this to d41b7e6d7a31d.

templates/web.letsencrypt.ssl.template.yml
```diff
@@ -19,6 +19,20 @@ run:
         LE_WORKING_DIR="${LETSENCRYPT_DIR}" ./acme.sh --upgrade --auto-upgrade
         LE_WORKING_DIR="${LETSENCRYPT_DIR}" ./acme.sh --set-default-ca  --server  letsencrypt
 
+        cat << EOF > /etc/nginx/conf.d/outlets/before-server/20-redirect-http-to-https.conf
```
andrewschleifer, line 22: In, for example, `samples/web_only.yml`, we tell operators to uncomment both `web.ssl.template.yml` and `web.letsencrypt.ssl.template.yml`.

However, the `configure-letsencrypt` and `configure-ssl` scripts are writing to the same outlet. So future changes to the latter are going to be mysteriously overwritten by the former.

templates/web.letsencrypt.ssl.template.yml
```diff
@@ -51,6 +64,12 @@ run:
         }
         EOF
 
+        if [ -f "/proc/net/if_inet6" ] ; then
```
andrewschleifer, line 67: Does this file exist if IPv6 is enabled but no interfaces have v6 addresses?

We want to listen on `::` even if there are current addresses, in case one appears in the future.

andrewschleifer (changes requested):

featheredtoast:
@andrewschleifer updated with feedback, let me know if that looks better.

templates/web.letsencrypt.ssl.template.yml
andrewschleifer: I don't actually know what the behavior is when IPv6 is enabled but not configured anywhere. Maybe that `if` test is OK?

If `listen [::]:80` does not error when IPv6 is not available that sure looks like the cleanest option. I'll try and figure out a test.

andrewschleifer (approved):

PR #994, opened 2025-08-14
Separate builders for build tools, drop some unused library dependencies

featheredtoast:
Separate out builders for thpoff, jemalloc and oxipng/jhead
Copy over the results of all libraries to reduce dependencies
for base image needing build tools.

drop extra -dev packages from base image as we only need the nginx runtime libs.

refactor parallel stop without parallel package

mount tmpfs for apt caches
remove unreferenced libtcmalloc

reorganize gem dependencies

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -78,13 +78,14 @@ RUN --mount=type=tmpfs,target=/var/log \
     cmake g++ pkg-config patch \
     libxslt-dev libcurl4-openssl-dev \
     libssl-dev libyaml-dev libtool \
-    libpcre3 libpcre3-dev zlib1g zlib1g-dev \
     libxml2-dev gawk \
     libreadline-dev anacron wget \
     psmisc whois brotli libunwind-dev \
     libtcmalloc-minimal4 cmake \
     pngcrush pngquant ripgrep poppler-utils \
     catdoc antiword \
+# nginx runtime dependencies
+    libpcre3 zlib1g \
 # imagemagick runtime dependencies
     ghostscript libjbig0 libtiff6 libpng16-16 libfontconfig1 \
     libwebpdemux2 libwebpmux3 libxext6 librsvg2-2 libgomp1 \
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -77,7 +77,7 @@ RUN --mount=type=tmpfs,target=/var/log \
     libxslt-dev libcurl4-openssl-dev \
     libssl-dev libyaml-dev libtool \
     libpcre3 libpcre3-dev zlib1g zlib1g-dev \
-    libxml2-dev gawk parallel \
+    libxml2-dev gawk \
     libreadline-dev anacron wget \
     psmisc whois brotli libunwind-dev \
     libtcmalloc-minimal4 cmake \
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -81,7 +81,6 @@ RUN --mount=type=tmpfs,target=/var/log \
     libxml2-dev gawk \
     libreadline-dev anacron wget \
     psmisc whois brotli libunwind-dev \
-    libtcmalloc-minimal4 cmake \
     pngcrush pngquant ripgrep poppler-utils \
     catdoc antiword \
 # nginx runtime dependencies
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -73,16 +73,23 @@ RUN --mount=type=tmpfs,target=/var/log \
   --mount=type=tmpfs,target=/var/cache/apt \
   --mount=type=tmpfs,target=/var/lib/apt \
     echo "debconf debconf/frontend select Teletype" | debconf-set-selections; \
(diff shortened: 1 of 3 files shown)


image/base/sbin/boot
```diff
@@ -5,15 +5,18 @@
...
 
   # give stuff a bit of time to finish
   sleep 0.1
 
   ORPHANS=`ps -eo pid | grep -v PID  | tr -d ' ' | grep -v '^1$'`
-  SHELL=/bin/bash parallel 'timeout 5 /bin/bash -c "kill {} && wait {}" || kill -9 {}' ::: $ORPHANS 2> /dev/null
+  for pid in $ORPHANS; do
+    (timeout 5 /bin/bash -c "kill $pid && wait $pid" 2>/dev/null || kill -9 $pid 2>/dev/null) &
+  done
+  wait
```
andrewschleifer, line 19: Do you know what the history is with this? Why are we killing ORPHANS in the first place? Unclean shutdown feels out of scope for this script and ought to be handled one level up.

image/base/Dockerfile
```diff
@@ -55,23 +70,34 @@ RUN echo 2.0.`date +%Y%m%d` > /VERSION
 RUN echo "deb http://deb.debian.org/debian ${DEBIAN_RELEASE}-backports main" > "/etc/apt/sources.list.d/${DEBIAN_RELEASE}-backports.list"
 
 RUN --mount=type=tmpfs,target=/var/log \
+  --mount=type=tmpfs,target=/var/cache/apt \
+  --mount=type=tmpfs,target=/var/lib/apt \
```
andrewschleifer, line 74: I don't know what the lifecycle of this container is. Are we potentially increasing the RAM usage for every Discourse here?

image/base/Dockerfile
```diff
@@ -55,23 +70,34 @@ RUN echo 2.0.`date +%Y%m%d` > /VERSION
...
-    libssl-dev libyaml-dev libtool \
-    libpcre3 libpcre3-dev zlib1g zlib1g-dev \
-    libxml2-dev gawk parallel \
-    libreadline-dev anacron wget \
-    psmisc whois brotli libunwind-dev \
-    libtcmalloc-minimal4 cmake \
+    gawk anacron wget \
+    psmisc whois brotli \
     pngcrush pngquant ripgrep poppler-utils \
     catdoc antiword \
```
andrewschleifer, line 83: Pulling and labelling the sections below is awesome. Should we do the same with the preceding as `# discourse runtime requirements`?

image/base/Dockerfile
```diff
@@ -127,24 +153,17 @@ RUN ln -s /usr/local/bin/magick /usr/local/bin/animate &&\
   ln -s /usr/local/bin/magick /usr/local/bin/stream &&\
   test $(magick -version | grep -o -e png -e tiff -e jpeg -e freetype -e heic -e webp | wc -l) -eq 6
 
-ADD install-jemalloc /tmp/install-jemalloc
-RUN /tmp/install-jemalloc
+COPY --from=thpoff_builder /usr/local/sbin/thpoff /usr/local/sbin
+COPY --from=jemalloc_builder /usr/lib/libjemalloc.so /usr/lib
+COPY --from=oxipng_builder /usr/local/bin/jhead /usr/local/bin
+COPY --from=oxipng_builder /usr/local/bin/oxipng /usr/local/bin
```
andrewschleifer: Bikeshed: Infrastructure convention is to use hyphens instead of underscores in these names.

andrewschleifer (changes requested):

image/base/sbin/boot
featheredtoast: I think it's related to how docker/launcher interacted with the containers previously, but I don't have history other than the note in git history:

> FIX: force all sub processes to stop on TERM, wait for them to finish
> This works around docker bug where containers will not stop

Change here attempted to remove parallel without rocking the boat too much on lifecycle.

image/base/Dockerfile
featheredtoast: No, this Dockerfile is only done by the builder for building the base dependencies image. All Discourse images /containers build or run later don't inherit a tmpfs from this.

The idea here is just to have an image that is clear of apt caches on downstream images.

We used to do this back in the day by doing a later `RUN rm-all-caches` and flattening image layers, but tmpfs is the way to achieve this now for directories we don't want saved in a final image.

*Cache mounting for apt is also given as an example on [Docker's docs page](https://docs.docker.com/build/cache/optimize/#use-cache-mounts) as well.

image/base/Dockerfile
featheredtoast: I've added it - Just for full transparency, I initially didn't label the first block above as it's less clear to me as to exactly what is a hard discourse requirement (I'm not so sure about a Discourse hard requirement on fping, gawk, psmisc, ripgrep, or whois for example).
Some are known for container boot (anacron, wget, gnupg, sudo, curl).
Others I know are Discourse requirements - pngcrush, pngquant, poppler-utils, catdoc, antiword.

It's possible there's crossover.

featheredtoast:
@andrewschleifer updated 👍

andrewschleifer (approved):

PR #1000, opened 2025-09-16
DEV: reduce image size by using web-only image for discourse/discourse images

featheredtoast:
Take advantage of web-only images to build discourse/discourse without postgres and redis invovled.

Now that the base image bump is in, we have a matching web-only version to build from


CvX (approved):
PR 1000 😌

PR #1048, opened 2026-03-12
Update to trixie

Supermathie:
* c742f106 FIX: use && instead of semicolons in Dockerfiles
* 8b230310 CLEANUP: remove yarn from base image
* fff21fee CLEANUP: use "&&\" everywhere instead of "&& \" for line continuance in RUN commands
* c4adf636 CLEANUP: remove libgconf-2-4 from discourse_test dockerfile
* d8b59d96 CLEANUP: Update from legacy dockerfile syntax
* a3f02929 DEV: base image: update to trixie


Supermathie:
still need to test the generated image, I think the workflow will build & push something I can try?

Supermathie:
Converted to draft as we're not yet ready to actually go to trixie due to the collation problem

… I think we could make all the other changes though 🤔

lightbulbjim (approved):

lightbulbjim (approved):

PR #1051, opened 2026-04-28
FEATURE: add antiword and catdoc

SamSaffron:
These are important for document extractions in case we need them


xfalcox (approved):

pfaffman:
I'm late to this party, but docx was introduced 20 years ago. Why are these legacy libraries that work only with .doc files important now?

SamSaffron:
The enterprise moves slowly

PR #1056, opened 2026-05-20
Drop removed bundler flags

davidtaylorhq:


image/discourse_dev/Dockerfile
```diff
@@ -66,7 +66,8 @@ USER discourse
 # Warm global bundle cache, then delete the compressed `cache/` versions (`/gem/` are enough)
 RUN --mount=type=bind,src=/repo,from=repo-fetcher,target=/tmp/discourse-clone,readwrite \
     cd /tmp/discourse-clone \
-    && bundle install --deployment \
+    && bundle config set deployment && \
```
Supermathie, line 69: should this not be `bundle config set deployment true` as per the error message?

fitzy101:
Might be easier to use `env BUNDLE_DEPLOYMENT=true bundle install ....` instead of having to use the call `bundle config set ...` or have the `--without test development` and `--path ..` in the argument list. 

e.g.

```
diff --git a/image/discourse_dev/Dockerfile b/image/discourse_dev/Dockerfile
index 9f9f4bf..fa42b65 100644
--- a/image/discourse_dev/Dockerfile
+++ b/image/discourse_dev/Dockerfile
@@ -66,7 +66,7 @@ USER discourse
 # Warm global bundle cache, then delete the compressed `cache/` versions (`/gem/` are enough)
 RUN --mount=type=bind,src=/repo,from=repo-fetcher,target=/tmp/discourse-clone,readwrite \
     cd /tmp/discourse-clone \
-    && bundle install --deployment \
+    && env BUNDLE_DEPLOYMENT=true bundle install \
     && rm -rf /home/discourse/.bundle/gems/ruby/*/cache/*
 
 # Warm global yarn cache
```

PR #1065, opened 2026-05-29
Push runtime deps image

featheredtoast:
Push an image with only Discourse runtime dependencies. Provides a base image for a finalized image that copies from a build image, resulting in much smaller image sizes.

Rename base_deps to base_build_deps to distinguish from base_runtime_deps*

base_build_deps* depends on base_runtime_deps* builds and can reuse the fresh runtime image build cache. The cache is guaranteed to be fresh as runtime images now handle dropping the docker build cache when built first.

Fix incorrect base_slim_stable_amd64 tag

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,10 +46,15 @@ jobs:
         with:
           fetch-depth: 1
 
-      - name: build deps image
+      - name: build runtime deps image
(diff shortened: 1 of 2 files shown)


ZogStriP (approved):

featheredtoast:
Implemented alongside #1088

PR #1066, opened 2026-05-31
DEV: use docker bake to build images

featheredtoast:
Replace the ruby auto_build.rb script with [docker bake](https://docs.docker.com/build/bake/introduction/)

Use build cache to push image tags directly with `docker bake --push` rather than specifying tags in `build.yml`

Unifies the way all dockerfile-based images are built.

Allows for image groups to be built/pushed at once.

Bake targets are able to handle target variants, and tags accordingly across all version and tags (version+arch), simplifying `build.yml`

Renames stable-* images to esr-*

Handle image building differences:

Test builds
---

Test images have two distinct bake targets: at build time (built at -base) and when pushing `test` jobs. base test build jobs are built from local images, before base images are pushed. Both amd64 and arm64 variants are built. The `test` job depends on images to get pushed and depends on the finalized pushed multiarch tags, and the test image is only built with the amd64 variant.

Dev builds
---

Update the way base templates are copied into the dev image - rather than an external process copying and mutating the template, copy the existing template straight from ../templates (via build context) and mutate within the dev dockerfile.

add explicit allow when bulding dev images

a dev build mounts up templates from the parent directory to setup postgres and redis from the template skeletons.

buildx may complain about insecure capabilities without this.

Previously copying and editing of templates was done in the ruby build script directly.

Parameters
---

Add parameters to many settings, allowing for the github actions to be run on forks, or against different environments. Eases testing outside of the main docker repositories when developing against discourse_docker.

Github forks would be able to build images themselves, allowing for personal docker repositories to be used, or potentially separate into official stage or testing environments by way of GitHub environments.

Add an option to skip tests via vars, making it easier to test changes.

All GitHub environments should then define following parameters:

* `AMD64_RUNNER` - Github actions runner for AMD64 image versions
* `ARM64_RUNNER` - Github actions runner for ARM64 image versions
* `BASE_IMAGE` - Docker repository for discourse base image builds
* `DEV_IMAGE` - Docker repository for discourse dev image builds
* `TEST_IMAGE` - Docker repository for discourse test image builds
* `SETUP_WIZARD_IMAGE` - Docker repository for discourse setup wizard image builds
* `WEB_ONLY_IMAGE` - Docker repository for discourse web-only image builds
* `DOCKERHUB_USERNAME` - docker registry username
* `DOCKERHUB_PASSWORD` - docker registry password
* `SKIP_TESTS` - Flag to skip tests. Set to 1 to skip full tests before images pushes.

Allows time for builds to complete on github-hosted runners.


featheredtoast:
This cleans up a lot of `build.yml` so it's easier to follow and make changes/add images. Makes it easier to run against forks or different environments for [easier development/testing](https://github.com/featheredtoast/discourse_docker/actions/runs/26715836738) and makes it much less complicated to change what images we're publishing or how they are tagged.

Introducing github environments:

| variable | test value | proposed discourse_docker value |
|---|---|---|
| `AMD64_RUNNER` | ubuntu-latest | cdck-linux-8-core |
| `ARM64_RUNNER` | ubuntu-24.04-arm | ubuntu-24.04-8core-arm |
| `BASE_IMAGE` | featheredtoast/discourse-base | discourse/base |
| `DEV_IMAGE` | featheredtoast/discourse-dev | discourse/discourse_dev |
| `SETUP_WIZARD_IMAGE` | featheredtoast/discourse-setup-wizard | discourse/setup-wizard |
| `SKIP_TESTS` | 1 | |
| `TEST_IMAGE` | featheredtoast/discourse-test | discourse/discourse_test |
| `WEB_ONLY_IMAGE` | featheredtoast/discourse-web-only | discourse/discourse |

Test branch: https://github.com/featheredtoast/discourse_docker/tree/refs/heads/docker-bake-test

Test repositories:
https://hub.docker.com/r/featheredtoast/discourse-base/tags
https://hub.docker.com/r/featheredtoast/discourse-setup-wizard/tags
https://hub.docker.com/r/featheredtoast/discourse-test/tags
https://hub.docker.com/r/featheredtoast/discourse-dev/tags

davidtaylorhq:
> Renames stable-* images to esr-*

That's a breaking change, right? Tbh, I don't think we should be publishing any tags like esr/stable/release. Instead, we should be maintaining tags for each currently-supported release, and then people should be responsible for picking the one they want and updating when they want. That way, it can work well with our new overlapping support policy.

So for the purposes of this PR, maybe we keep it named `stable`, and then work out that kind of change as a future PR?

.github/workflows/build.yml
```diff
@@ -1,7 +1,5 @@
 on:
   push:
-    branches:
-      - main
```
davidtaylorhq, line 4: We should keep the `main` restriction, right?

.github/workflows/build.yml
```diff
@@ -26,285 +29,215 @@ jobs:
           echo "timestamp=$timestamp" >> $GITHUB_OUTPUT
 
   base:
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == format('refs/heads/{0}', vars.IMAGE_PUSH_BRANCH)) }}
```
davidtaylorhq: Is this working correctly? seems like all the CI steps were skipped on the most recent push to this PR 👀

.github/workflows/build.yml
```diff
@@ -26,285 +29,215 @@ jobs:
...
-          docker run --rm -e RUBY_ONLY=1 -e USE_TURBO=1 -e SKIP_PLUGINS=1 -e SKIP_LINT=1 -e DISCOURSE_TURBO_RSPEC_RETRY_AND_LOG_FLAKY_TESTS=1 discourse/discourse_test:build_${{ matrix.arch }}
+          docker run --rm -e RUBY_ONLY=1 -e USE_TURBO=1 -e SKIP_PLUGINS=1 -e SKIP_LINT=1 -e DISCOURSE_TURBO_RSPEC_RETRY_AND_LOG_FLAKY_TESTS=1 ${TEST_REPO}:build-${{ matrix.arch }}
 
-      - name: build & tag dev image for `main` branch
-        working-directory: image
+      - name: Build and push dev image
         run: |
-          ruby auto_build.rb discourse_dev_${{ matrix.arch }}
-          docker tag discourse/discourse_dev:build_${{ matrix.arch }} discourse/discourse_dev:${{ env.TIMESTAMP }}-${{ matrix.arch }}
+          docker buildx bake dev --push --allow=fs.read=../templates
```
davidtaylorhq: Should this step be gated on `main`? Otherwise we're gonna be pushing images from PR CI?

image/README.md
```diff
@@ -2,22 +2,49 @@
...
 ## Building new images
 
-To build a new image, just run `ruby auto_build.rb image-name`. The build process will build a local image with a predefined tag.
+To build a new image, just run `docker bake {target}`. The build process will build a local image with predefined tags.
 
-Images and tag names are defined [here](https://github.com/discourse/discourse_docker/blob/main/image/auto_build.rb#L6-L88).
+Run `docker bake --list targets` to see the list of valid targets.
 
-> **A note about --squash**: By default we squash the images we serve on Docker Hub. You will need to [enable experimental features](https://github.com/docker/docker-ce/blob/master/components/cli/experimental/README.md) on your Docker daemon for that.
+See [docker bake](docs.docker.com/build/bake) for more details.
```
davidtaylorhq:
```suggestion
See [docker bake](https://docs.docker.com/build/bake) for more details.
```
(otherwise it renders as a relative link)

.github/workflows/build.yml
```diff
@@ -26,285 +29,215 @@ jobs:
...
+            ${SETUP_WIZARD_REPO}:release-amd64 \
+            ${SETUP_WIZARD_REPO}:release-arm64
 
   test:
-    runs-on: cdck-linux-8-core
-    timeout-minutes: 30
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == format('refs/heads/{0}', vars.IMAGE_PUSH_BRANCH)) }}
+    runs-on: ${{ vars.AMD64_RUNNER }}
+    timeout-minutes: 60
     needs: base
```
davidtaylorhq: Not related to your PR, but we should probably make this:


```suggestion
    needs: push_multi_arch_manifests
```
otherwise there's a race condition where it'll test the older version of the images.

(which also makes me think... maybe this job shouldn't run at all on PRs? Kinda silly to test the released images, and not the ones built from the current PR...)

image/README.md
```diff
@@ -2,22 +2,49 @@
 
 ## Building new images
 
-To build a new image, just run `ruby auto_build.rb image-name`. The build process will build a local image with a predefined tag.
+To build a new image, just run `docker bake {target}`. The build process will build a local image with predefined tags.
```
davidtaylorhq:
```suggestion
To build a new image, just run `docker buildx bake {target}`. The build process will build a local image with predefined tags.
```
?

And the same for all the other references in the readme?

davidtaylorhq:
I asked claude to make a diff of the published tags before/after this change. Seems like there are a lot of new ones being published:

```diff
 discourse/base

  # --- per-arch tags ---
- 2.0.T-slim-{arch}
+ 2.0.T-slim-main-{arch}          # renamed: gained -main qualifier
+ 2.0.T-slim-esr-{arch}           # NEW (old built stable slim but never pushed it)
+ slim-main-{arch}                # NEW (plain, non-timestamped)
+ slim-esr-{arch}                 # NEW
+ web-only-main-{arch}            # NEW (plain)
+ web-only-esr-{arch}             # NEW (plain)
+ release-main-{arch}             # NEW (plain)
+ release-esr-{arch}              # NEW (plain)
- 2.0.T-web-only-stable-{arch}
+ 2.0.T-web-only-esr-{arch}       # rename stable→esr
- 2.0.T-stable-{arch}
+ 2.0.T-esr-{arch}                # rename stable→esr
  2.0.T-web-only-main-{arch}      # unchanged
  2.0.T-main-{arch}               # unchanged
  aarch64                         # unchanged

  # --- multi-arch manifests (user-facing) ---
- web-only-stable   
+ web-only-esr                    # rename
- 2.0.T-web-only-stable
+ 2.0.T-web-only-esr              # rename
- 2.0.T-stable
+ 2.0.T-esr                       # rename
- release-stable
+ release-esr                     # rename
  slim                            # unchanged
  2.0.T-slim                      # unchanged
  web-only                        # unchanged
  2.0.T-web-only                  # unchanged
  release                         # unchanged
  2.0.T                           # unchanged

discourse/discourse_dev

+ release-{arch}                  # NEW (plain, non-timestamped per-arch)
  T-{arch}                        # unchanged
  release                         # unchanged (manifest)
  T                               # unchanged (manifest)

discourse/setup-wizard — no change

  release-{arch}                  # unchanged
  release                         # unchanged (manifest)

discourse/discourse_test — no change

  slim, slim-browsers, release    # unchanged
```
is that intentional?

image/README.md
featheredtoast: As I understand it `docker bake` is an alias for `docker buildx bake` (similar to `docker build` and friends). I documented the friendly name by default.

.github/workflows/build.yml
featheredtoast: Good catch - updated 👍

.github/workflows/build.yml
featheredtoast: The discourse_docker env vars need to be updated for the proposed github actions workflow to run properly. I can update those in the repo so the workflows can run properly.

.github/workflows/build.yml
featheredtoast: It's intentional. I'm parameterizing `main` to get this repository easier to test, and skipping workflows that do not match `vars.IMAGE_PUSH_BRANCH` (IE, `main`).

There's unfortunately no way to do something like:
```
push:
  branches:
    - ${{ vars.IMAGE_PUSH_BRANCH }}
```

so the if blocks are the best we have.

---

Doing this allows us to have a future where we can build/push in different parameters, such as setting up test docker repos to iterate on changes outside of the `main` branch

featheredtoast:
For the new branches, there's no good way for bake to omit pushing branches in the list unfortunately, so I'm proposing pushing those out for sake of simplicity.

That said, it's not ideal. Would really be nice to have a way to delete branches programmatically though. Docker's experimental [hub-tool](https://github.com/docker/hub-tool) was able to do it, but was archived this year, so maybe there's traction to get it added at some point.

Some thoughts to do tagging more elegantly:
* push to private repo, pull, retag, and push publicly
* export after load, upload individual build artifacts, final job that imports all docker image tarballs, tags and pushes
* rewrite to use [multiplatform](https://docs.docker.com/build/ci/github-actions/multi-platform/#with-bake) builds - this *might* be possible now that we're on a single bake file, I just haven't yet dove in. Would be nice if it works, but looks like we wouldn't be able to hook + run tests before pushing.

featheredtoast:
Probably the most attractive option re: consolidating tag names is to cache to another image registry (ghcr? internal?) It also means we aren't tightly dependent on github-specific bits.

Then we can build separately, cache the run, and then when we're ready to push, do that for all images. There is also a lot of opportunity for parallel builds, reusing the cache. I'll get a POC out soon.

.github/workflows/build.yml
featheredtoast: After mulling it over, I've updated it so test branches only need to update the `main` restriction to test different branches -- before you needed to change it in a bunch of different places

featheredtoast:
OK so some rework for quite a few improvements:

* updated how we're pushing - we can push by digest to omit all the -arch tags from being pushed and I've done so here. (`:aarch64` is still pushed for backwards compatiblity)
* kept `main` as push taget, but reworked how pushes are triggered, the idea here is to make it easy to test on different branches. Test branches can just update the target branch at the top and push temporarily, rather than needing to search/replace `s/main/my-test-branch`
* test is now built like the other images, and is pushed/tagged with the rest as a multiarch image, meaning the right image is pushed, and also it does not need to be built twice (as is the case today)

I've also went ahead and updated the repository variables so the job can run now.

image/README.md
featheredtoast: Second thought, perhaps it is better to document docker buildx bake if one is using different builders. Updated.

.github/workflows/build.yml
featheredtoast: I've addressed this by removing this particular job and rolling the build/push into the rest of the builds so the right image is getting pushed/tagged.

I think this was due to a limitation in the FROM line, where docker will helpfully not let you use a remote declared image without checking for updates first. I've updated this so FROM takes a context instead (where the context is a bake target).

This means:
* faster jobs, just building test once instead of twice per run
* more reliable builds of test (your race condition is resolved)
* more flexible test images with added multiarch support

davidtaylorhq (approved):
I opened a PR for a possible improvement, but not a blocker: https://github.com/discourse/discourse_docker/pull/1079

Otherwise looks good to me! 🚀

featheredtoast:
@davidtaylorhq I'm gonna merge as-is, and we can iterate on golfing/DRYing a bit later - there's some subtle things that happen with changing tags on docker push that we'll need to be careful with. Definitely want to continue on that (and things like build caching), but I think it's best to get these changes in now as-is.

PR #1067, opened 2026-06-02
Remove deprecated sshd and web.ipv6 templates

ducks:
Both files were marked `# TODO(2026-01-01): Remove this file` for over five months. They emit deprecation warnings when included but no longer provide any functionality:

- sshd was removed as a supported feature
- IPv6 is now enabled by default when available


featheredtoast (approved):

andrewschleifer:
If we delete the file, what does that look like for somebody that has it included? Is there a reasonable error message?

PR #1069, opened 2026-06-09
Update web.china.template.yml

AkarinLiu:
update from ghproxy.com to gh-proxy.com


andrewschleifer:
https://gh-proxy.com is not loading, who runs the service?

PR #1070, opened 2026-06-09
Drop ghostscript dependency

davidtaylorhq:
We don't need this, and it adds unnecessary vulnerability surface-area


ZogStriP (approved):
🔥 👻📜

KrisKotlarek (approved):

PR #1073, opened 2026-06-11
DEV: pin external dependencies and update

SamSaffron:
Vendor NodeSource and Yarn signing keys instead of fetching them during builds, and install pups from a verified local gem.

Pin git-sourced build inputs to immutable commits and refresh image dependencies including ImageMagick, jemalloc, oxipng, and Redis. Use HTTPS for Redis downloads and avoid installing the extra jemalloc 5 library on non-arm64 builds.


image/base/install-jemalloc
```diff
@@ -12,10 +12,10 @@ if uname -m | grep -qi 'aarch64'; then
   mkdir /jemalloc-new
   cd /jemalloc-new
 
-  wget -q https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2
-  sha256sum jemalloc-5.3.0.tar.bz2
-  echo "2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa jemalloc-5.3.0.tar.bz2" | sha256sum -c
-  tar --strip-components=1 -xjf jemalloc-5.3.0.tar.bz2
+  wget -q https://github.com/jemalloc/jemalloc/releases/download/5.3.1/jemalloc-5.3.1.tar.bz2
+  sha256sum jemalloc-5.3.1.tar.bz2
```
fitzy101, line 16: Is this line leftover from someone wanting to read the checksum? (I realise this was here already, but it's probably not needed?)

fitzy101 (approved):
Hardening these installs LGTM

image/base/install-jemalloc
SamSaffron: yeah I think we can drop it... we dont need 5.3.1

SamSaffron: oh it is needed for ARM

PR #1076, opened 2026-06-13
DEV: install VIPS from source

SamSaffron:
Debian vips depends on image magick core, by using it we end up shipping 2 versions of image magick in our base image.

This compiles vips library without any image magick sidestepping the problem.

Additionally ensures ghostscript is never installed in our image.


image/base/install-vips
```diff
@@ -0,0 +1,113 @@
...
+  --buildtype=release \
+  --auto-features=disabled \
+  -Ddeprecated=false \
+  -Dexamples=false \
+  -Dcplusplus=false \
+  -Dppm=false \
+  -Danalyze=false \
+  -Dradiance=false \
+  -Dmodules=disabled \
+  -Dmagick=disabled \
```
SamSaffron: this is how we get image magick free build

image/base/Dockerfile
```diff
@@ -68,12 +94,20 @@ RUN groupadd --gid 104 postgres &&\
     groupadd --gid 1000 discourse &&\
     useradd --uid 1000 --gid 1000 -m --shell /bin/bash discourse
 
-RUN echo 2.0.`date +%Y%m%d` > /VERSION
+# VERSION_DATE busts this layer (and the runtime apt layers below it) once a
+# day even when building from a registry cache; the compile-heavy builder
```
davidtaylorhq: We only build the image once per day... so what does the cache help with? 🤔

SamSaffron: may have to not bust it ...

davidtaylorhq:
Discussed in chat - we're gonna split this out into separate proposals

SamSaffron:
should we go with this @davidtaylorhq  ?

davidtaylorhq (approved):

PR #1077, opened 2026-06-15
FIX: fastly template should be teaching nginx what to trust

Supermathie:
The idea of replacing proxy_add_x_forwarded_for is ghastly - we have the
ability to teach nginx how to discern the correct values for the actual end
user IP instead of telling it to pass through the header to Discourse.

This means the server-side nginx logs will be correct, plus allows for more
complex setups.

### DIFF
--- a/templates/fastly.template.yml
+++ b/templates/fastly.template.yml
@@ -1,6 +1,11 @@
 run:
-  - replace:
-     filename: "/etc/nginx/conf.d/discourse.conf"
-     from: "$proxy_add_x_forwarded_for"
-     to: "$http_fastly_client_ip"
-     global: true
+  - file:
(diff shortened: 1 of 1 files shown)


fitzy101 (approved):

templates/fastly.template.yml
```diff
@@ -1,6 +1,11 @@
...
-     to: "$http_fastly_client_ip"
-     global: true
+  - file:
+      path: /etc/nginx/conf.d/outlets/server/real-ip-header.conf
+      chmod: 644
+      contents: |
+        real_ip_header fastly-client-ip;
+  - exec:
+      cmd:
+        - apt-get update && apt-get install -y jq && apt-get clean
```
Supermathie, line 9: I feel a *little* bad about this, but after we move to trixie we can pull `yq-go` into the base image for tasks like this

PR #1079, opened 2026-06-16
DEV: drive multi-arch manifests from a single digest map

davidtaylorhq:
Previously each matrix leg hand-extracted its image digests into many named job outputs (`build.yml` re-listed 22), and the manifest jobs re-listed every variant again as `docker buildx imagetools create` blocks — parallel copies of the same variant matrix to keep in sync.

This change makes each arch leg emits one compact `{ "<target>": "<digest>" }` JSON map as a single job output (keyed by the names `docker-bake.hcl` generates). The manifest job merges the per-arch maps and creates every published manifest from a compact looped table via shared `ref`/`manifest` helpers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,15 +81,13 @@ jobs:
         run: |
           docker run --rm -e RUBY_ONLY=1 -e USE_TURBO=1 -e SKIP_PLUGINS=1 -e SKIP_LINT=1 -e DISCOURSE_TURBO_RSPEC_RETRY_AND_LOG_FLAKY_TESTS=1 ${TEST_IMAGE}:release-${{ matrix.arch }}
 
-      - name: Build and push test image
+      - name: Build and push images
         if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
         run: |
-          docker buildx bake test --set="*.tags=${TEST_IMAGE}" --set="*.output=type=registry,push-by-digest=true" --metadata-file=/tmp/test.json
-
-      - name: Build and push dev image
-        if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -81,15 +81,13 @@ jobs:
...
         if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
         run: |
-          docker buildx bake test --set="*.tags=${TEST_IMAGE}" --set="*.output=type=registry,push-by-digest=true" --metadata-file=/tmp/test.json
-
-      - name: Build and push dev image
-        if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
-        run: |
-          docker buildx bake dev --set="*.tags=${DEV_IMAGE}" --set="*.output=type=registry,push-by-digest=true" --allow=fs.read=../templates --metadata-file=/tmp/dev.json
+          docker buildx bake all \
+            --set="*.output=type=registry,push-by-digest=true" \
```
featheredtoast, line 88: One thing that may still need some extra setting here is tags, tags need to be *just* the image name, with no 'tagged' part (EG, `discourse/base` instead of `discourse/base:release`) to be able to push by digest. Because Docker-isms 😭 

The bake file could be made to be more GHA-friendly by making the tags more friendly to this, or maybe that can be a matrix, but my default MO was make it the most cli-intuitive to build images, and do the weird fiddly GHA-related bits like tagging for push-by-digest in GHA. But we should just make it more CI friendly from the get-go. There's also ways of making a bake override file just for CI instead. 🤷 

Anyway, I think this is a good direction - I'm erring on merging #1066 as-is for now and we can follow-up with golfing it further.

ZogStriP (approved):

PR #1087, opened 2026-07-13
FEATURE: Add Postgres 18 template

lightbulbjim:
This allows bootstrapping new containers as well as upgrading old running containers to version 18. Note that upgrades have only been tested from version 15.

In the future Postgres 18 will become the default and upgrade documentation will be made available at that time.

### DIFF
--- a/templates/postgres.18.template.yml
+++ b/templates/postgres.18.template.yml
@@ -1,0 +1,292 @@
+params:
+  db_synchronous_commit: "off"
+  db_shared_buffers: "256MB"
+  db_work_mem: "10MB"
+  db_default_text_search_config: "pg_catalog.english"
+  db_name: discourse
+  db_user: discourse
+  db_checkpoint_segments: 6
+  db_logging_collector: off
+  db_log_min_duration_statement: 100
+
+hooks:
+  before_code:
+    - replace:
+        filename: /etc/service/unicorn/run
+        from: "# postgres"
+        to: |
+          sv start postgres || exit 1
+
+run:
+  - file:
+      path: /etc/service/postgres/run
+      chmod: "+x"
+      contents: |
+        #!/bin/sh
+        exec 2>&1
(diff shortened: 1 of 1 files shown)


pfaffman:
This follows the same rebuild twice to upgrade flow as the last several postgres upgrades? 

I'll update my scripts in a day or two and try it on a bunch of sites

lightbulbjim:
Yes, the same double rebuild workflow. Let us know if you encounter any problems.

There's no PG18 server in the base image yet so you'll need to make that change yourself for now. It will be added by the time PG18 becomes the default but for now we just need the template to consume internally.

pfaffman:
Oh! So we don't get have the double rebuild workflow, but if I want to upgrade my external postgres servers to 18, they'll work with the default web image. 

Sorry I missed that

lightbulbjim:
Sorry, I should have elaborated more. It's a little complicated right now while we work towards making PG18 the new default.

The `discourse-web-only` variant of the image contains both PG15 and PG18 clients and is intended to run with a standalone Postgres server.

The `discourse-release` variant is derived from `discourse-web-only` and adds the PG15 server. This is intended for running everything in a single container but can also be used with a separate Postgres server (which is how we use it internally).

The PG18 template added by this PR isn't referenced by anything in this repo yet. We have some internal tooling which will consume the template and unblock the PG18 upgrade on our hosted platform. At some point during or after that upgrade we'll replace the PG15 server with PG18 in the release image and templates and then anyone who pulls the latest version of `discourse_docker` will be upgraded.

If you want to test it now you can do so by updating the image referenced in `launcher` and the templates to one which includes the PG18 server, and using this new Postgres template instead of the default. That will use the double rebuild workflow to upgrade an existing PG15 container.

We are considering options to simplify some of this in the future. Right now the upgrade process is a bit complicated because of the way we use this repo internally.

fitzy101 (approved):

PR #1088, opened 2026-07-21
Build cache

featheredtoast:
Add build cache to builds, cache to ghcr. Cache by current date.

Refactor bake config: remove references to arch, and split up bake configuration to be able to compose image building into parts:
* base bake
* push
* cache-read
* cache-write

Allows for cache-read on PR, cache-read+write on main builds, push after tests on main.

For caching, we unfortunately explicitly need to run docker buildx bake TARGET for EVERY target individually. Targeting a group or multiple targets at the same time results in cache misses with seemingly no rhyme or reason - I'm guessing that despite each target having a separate cache-from/cache-to config, the cache-* sections are all grouped together across the entire build. Docker build caches apparently take the first matching cache, and discard the rest despite a cache-hit being possibly available in a different cache.

Allow cache to be flushed by searching for untagged images on ghcr, and delete

Fix test build to build every step from slim, previously the test release would build from base:release. Now it builds from the previous test build step

Publish runtime-deps-only image

prefer COPY to ADD as COPY for explicitly for copying docker contexts.

### DIFF
--- a/image/docker-bake.cache.hcl
+++ b/image/docker-bake.cache.hcl
@@ -1,0 +1,88 @@
+variable "CACHE_IMAGE" {
+  default = ""
+  validation {
+    condition = CACHE_IMAGE != ""
+    error_message = "The variable 'CACHE_IMAGE' must not be empty."
+  }
+}
+
+variable "ARCH" {
+  deafault = ""
+  validation {
+    condition = ARCH != ""
+    error_message = "The variable 'ARCH' must not be empty."
+  }
+}
+
+target "base-runtime-deps" {
+  cache-from = [{
+    type = "registry"
+    ref = "${CACHE_IMAGE}:cache-runtime-deps-${ARCH}"
+  }]
+}
+target "base-build-deps" {
+  cache-from = [{
+    type = "registry"
+    ref = "${CACHE_IMAGE}:cache-build-deps-${ARCH}"
+  }]
+}
+target "base-slim" {
+  name = "base-slim-${branch}"
+  matrix = {
+    branch = ["main", "stable"]
+  }
+  cache-from = [{
+    type = "registry"
+    ref = "${CACHE_IMAGE}:cache-slim-${branch}-${ARCH}"
+  }]
+}
+target "base-web-only" {
+  name = "base-web-only-${branch}"
+  matrix = {
+    branch = ["main", "stable"]
+  }
+  cache-from = [{
+    type = "registry"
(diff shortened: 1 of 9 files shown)


image/docker-bake.cache.hcl
```diff
@@ -0,0 +1,88 @@
+variable "CACHE_IMAGE" {
+  default = ""
+  validation {
+    condition = CACHE_IMAGE != ""
+    error_message = "The variable 'CACHE_IMAGE' must not be empty."
+  }
+}
+
+variable "ARCH" {
+  deafault = ""
```
ducks: small typo here

image/docker-bake.hcl
```diff
@@ -22,137 +14,108 @@ variable "DEV_IMAGE" {
...
   matrix = {
-    arch = ARCH_ARRAY
     branch = ["main", "stable"]
   }
   context = "./base"
-  tags = ["${BASE_IMAGE}:slim-${branch}-${arch}"]
+  tags = ["${BASE_IMAGE}:slim-${branch}"]
   target = "discourse-slim"
-  platforms = ["linux/${arch}"]
   args = {
```
ducks, line 47: should DATESTAMP be passed here and to the remaining targets?

featheredtoast: My understanding is it's not necessary: because the base-runtime-deps is a dependency for all later targets, once the runtime-deps image is re-calculated, all subsequent layers+dependent images are also invalidated

ducks: Hmm, I don’t know if Bake establishes a dependency between these targets. Each target builds the Dockerfile independently, and `bake --print` shows `base-runtime-deps` receiving `DATESTAMP` while base-slim-main does not.

I wonder if the inconsistent `DATESTAMP` could also explain some of the grouped-target cache misses? A grouped invocation currently builds the shared runtime stage with today’s value for `base-runtime-deps`, but with the default 0 for the remaining targets, producing two different cache graphs.

featheredtoast: it's possible? I can put them in and we can give it a shot.

The cache misses I observed previously I'm fairly certain were before the introduction of the DATESTAMP, but I'm willing to try anything at this point.

featheredtoast: I've made updates for DATESTAMP, but still seeing cache misses when I'm testing with bake groups.

ducks (approved):

PR #1093, opened 2026-07-28
FEATURE: automatically retry after disk is clean

pyxfluff:
Currently when performing a docker cleanup when your server is full, the script will exit.

This fixes that logic by asking the user if they would like to continue the rebuild when a disk cleanup is successful. The script will re-evaluate the disk usage and prompt the user to continue the rebuild instead of exiting.

### DIFF
--- a/launcher
+++ b/launcher
@@ -300,23 +300,42 @@ check_prereqs() {
   fi
 
   # 7. enough space for the bootstrap on docker folder
-  folder=`$docker_path info --format '{{.DockerRootDir}}'`
+  folder=$("$docker_path" info --format '{{.DockerRootDir}}')
   safe_folder=${folder:-/var/lib/docker}
-  if [[ -d $safe_folder && $(stat -f --format="%a*%S" $safe_folder)/1024**3 -lt 5 ]] ; then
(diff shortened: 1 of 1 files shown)


launcher
```diff
@@ -300,23 +300,42 @@ check_prereqs() {
...
-      if [[ $REPLY =~ ^[Yy]$ ]]
-      then
+
+      if [[ $REPLY =~ ^[Yy]$ ]]; then
         $docker_path container prune --force --filter until=24h >/dev/null
         $docker_path image prune --all --force --filter until=24h >/dev/null
-        echo "If the cleanup was successful, you may try again now"
+        echo "Finished cleaning"
+        if check_disk; then
+          read -p "Continue rebuild? (Y/N) "
```
ZogStriP: Nitpick: but we use uppercase to denote the default choice.

```suggestion
          read -p "Continue rebuild? (y/N) "
```

ZogStriP (approved):

launcher
pyxfluff: Thanks, learned something new today 😅

ZogStriP (approved):

PR #1098, opened 2026-07-30
FEATURE: Upgrade to PostgreSQL 18

lightbulbjim:
As part of this upgrade we are switching the default DB locale to C.UTF-8 using Postgres' builtin provider. This locale uses code point values for collation and is not dependent on the OS' glibc, preventing index corruption on OS upgrades.

Note the differences pre- and post-upgrade:

```
discourse=> show server_version;
         server_version          
---------------------------------
 15.18 (Debian 15.18-1.pgdg12+1)
(1 row)

discourse=> \l
                                                      List of databases
   Name    |  Owner   | Encoding | Locale Provider |   Collate   |    Ctype    | Locale | ICU Rules |   Access privileges    
-----------+----------+----------+-----------------+-------------+-------------+--------+-----------+------------------------
 discourse | postgres | UTF8     | libc            | en_US.UTF-8 | en_US.UTF-8 |        |           | =Tc/postgres          +
           |          |          |                 |             |             |        |           | postgres=CTc/postgres +
           |          |          |                 |             |             |        |           | discourse=CTc/postgres
 postgres  | postgres | UTF8     | libc            | en_US.UTF-8 | en_US.UTF-8 |        |           | 
 template0 | postgres | UTF8     | libc            | en_US.UTF-8 | en_US.UTF-8 |        |           | =c/postgres           +
           |          |          |                 |             |             |        |           | postgres=CTc/postgres
 template1 | postgres | UTF8     | libc            | en_US.UTF-8 | en_US.UTF-8 |        |           | =c/postgres           +
           |          |          |                 |             |             |        |           | postgres=CTc/postgres
(4 rows)
```

```
discourse=> show server_version;
        server_version         
-------------------------------
 18.4 (Debian 18.4-1.pgdg12+1)
(1 row)

discourse=> \l
                                                  List of databases
   Name    |  Owner   | Encoding | Locale Provider | Collate |  Ctype  | Locale  | ICU Rules |   Access privileges    
-----------+----------+----------+-----------------+---------+---------+---------+-----------+------------------------
 discourse | postgres | UTF8     | builtin         | C.UTF-8 | C.UTF-8 | C.UTF-8 |           | =Tc/postgres          +
           |          |          |                 |         |         |         |           | postgres=CTc/postgres +
           |          |          |                 |         |         |         |           | discourse=CTc/postgres
 postgres  | postgres | UTF8     | builtin         | C.UTF-8 | C.UTF-8 | C.UTF-8 |           | 
 template0 | postgres | UTF8     | builtin         | C.UTF-8 | C.UTF-8 | C.UTF-8 |           | =c/postgres           +
           |          |          |                 |         |         |         |           | postgres=CTc/postgres
 template1 | postgres | UTF8     | builtin         | C.UTF-8 | C.UTF-8 | C.UTF-8 |           | =c/postgres           +
           |          |          |                 |         |         |         |           | postgres=CTc/postgres
(4 rows)
```

In order to switch the DB locale the upgrade process now runs a dump/restore instead of just calling pg_upgrade on the data directory. The downside of this approach is that during the upgrade we need to store three copies of the data (the old and new data directories plus the dump), so more disk space is required. It is possible to get around the storage requirement by streaming the output of pg_dump into psql, but this process is slow compared to the multithreading available when dumping to and restoring from disk.

As with other upgrades the old database files are left untouched at /shared/postgres_data_old in case of upgrade failure. These can either be deleted after a successful upgrade or retained for posterity.

Note we are still installing the PG15 client alongside PG18 for now. This allows the app to backup/restore to both versions on our hosted platform during the upgrade period.

The Meta support topic referenced in the upgrade script will be finalised and published when the PR is merged.


templates/postgres.18.template.yml
```diff
@@ -129,18 +129,34 @@ run:
           cp -pr /etc/postgresql/${PG_MAJOR_OLD}/main/* /shared/postgres_data
           echo  >> /shared/postgres_data/postgresql.conf
           echo "data_directory = '/shared/postgres_data'" >> /shared/postgres_data/postgresql.conf
+        
+          rm -fr /shared/postgres_dump
+          install -d -m 0755 -o postgres -g postgres /shared/postgres_dump
+        
+          su postgres -c "/usr/lib/postgresql/${PG_MAJOR_OLD}/bin/pg_ctl start -D /shared/postgres_data -o '-p 5432'"
+          su postgres -c "/usr/lib/postgresql/${PG_MAJOR_OLD}/bin/pg_dumpall --globals-only --clean -f /shared/postgres_dump/globals.sql"
+          su postgres -c "/usr/lib/postgresql/${PG_MAJOR_OLD}/bin/pg_dump -Fd -j4 discourse -f /shared/postgres_dump/discourse"
```
fitzy101: I wonder if we should attempt to calculate an appropriate `-j4` value. If this is run on a 2-CPU server, we may run into significant CPU contention. Conversely, if running this on a 16-CPU server, we could more aggressively parallelise.

Maybe `nproc - 1` or something?

Not sure if it's worth the additional effort, will leave it up to you.

fitzy101 (approved):

templates/postgres.18.template.yml
```diff
@@ -129,18 +129,34 @@ run:
           cp -pr /etc/postgresql/${PG_MAJOR_OLD}/main/* /shared/postgres_data
           echo  >> /shared/postgres_data/postgresql.conf
           echo "data_directory = '/shared/postgres_data'" >> /shared/postgres_data/postgresql.conf
+        
+          rm -fr /shared/postgres_dump
+          install -d -m 0755 -o postgres -g postgres /shared/postgres_dump
+
```
fitzy101, line 151: This is a critical section, we need error handling to ensure all of the commands succeed.

We should, at the least, toggle `set -o errexit` for sections that need it.

fitzy101 (dismissed):

templates/postgres.18.template.yml
lightbulbjim: Hmm, interesting. In general if any of the commands fail the restore will fail, and that is handled by the `$SUCCESS` var. The state is deliberately not rolled back on failure to allow investigation.

For these two lines in particular I'm thinking we probably only need to ensure the `rm` succeeds. If we have lingering data from a previous upgrade in the directory then we might restore the wrong thing.

`errexit` won't help us because `rm -f` always exits 0. I'll add an explicit directory check after it's removed.

Once we know the directory has been well and truly deleted the `install` should be safe as is. If it fails then the dump will fail, which will bubble up to the restore failing and setting `$SUCCESS` to false. No data loss will occur and the user will get the failure notification.

templates/postgres.18.template.yml
lightbulbjim: Good idea, done.

templates/postgres.18.template.yml
lightbulbjim: Actually on second thought it's just going to be safer to abort if any step fails. I'll do that.

lightbulbjim: The upgrade now fails on any error. The failure message also contains the failing command so we have a breadcrumb to follow if users report problems.

Traditionally we would just report a line number but we're several layers down the YAML rabbit hole so that is not so convenient.

lightbulbjim: Example failure (synthetic error):

```
waiting for server to start....2026-07-31 05:34:19.021 UTC [1147] LOG:  starting PostgreSQL 15.18 (Debian 15.18-1.pgdg12+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 12.2.0-14+deb12u1) 12.2.0, 64-bit
2026-07-31 05:34:19.021 UTC [1147] LOG:  listening on IPv6 address "::1", port 5432
2026-07-31 05:34:19.021 UTC [1147] LOG:  listening on IPv4 address "127.0.0.1", port 5432
2026-07-31 05:34:19.023 UTC [1147] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2026-07-31 05:34:19.025 UTC [1150] LOG:  database system was interrupted; last known up at 2026-07-31 05:31:59 UTC
2026-07-31 05:34:19.176 UTC [1150] LOG:  database system was not properly shut down; automatic recovery in progress
2026-07-31 05:34:19.178 UTC [1150] LOG:  redo starts at 0/2C06C08
2026-07-31 05:34:19.180 UTC [1150] LOG:  invalid record length at 0/2C38620: wanted 24, got 0
2026-07-31 05:34:19.180 UTC [1150] LOG:  redo done at 0/2C385E8 system usage: CPU: user: 0.00 s, system: 0.00 s, elapsed: 0.00 s
2026-07-31 05:34:19.183 UTC [1148] LOG:  checkpoint starting: end-of-recovery immediate wait
2026-07-31 05:34:19.191 UTC [1148] LOG:  checkpoint complete: wrote 117 buffers (0.7%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.003 s, sync=0.006 s, total=0.009 s; sync files=83, longest=0.003 s, average=0.001 s; distance=198 kB, estimate=198 kB
2026-07-31 05:34:19.193 UTC [1147] LOG:  database system is ready to accept connections
 done
server started
pg_ctl: directory "/shared/postgres_dat" does not exist
-------------------------------------------------------------------------------------
UPGRADE OF POSTGRES FAILED
Failed on: su postgres -c "/usr/lib/postgresql/${PG_MAJOR_OLD}/bin/pg_ctl stop -D /shared/postgres_dat"

Please visit https://meta.discourse.org/t/postgresql-18-update/406194 for support.

You can run ./launcher start app to restart your app in the meantime
-------------------------------------------------------------------------------------


FAILED
--------------------
Pups::ExecError: if [ -f /root/install_postgres ]; then
  /root/install_postgres && rm -f /root/install_postgres
elif [ -e /shared/postgres_run/.s.PGSQL.5432 ]; then
  socat /dev/null UNIX-CONNECT:/shared/postgres_run/.s.PGSQL.5432 || exit 0 && echo postgres already running stop container ; exit 1
fi
 failed with return #<Process::Status: pid 18 exit 1>
Location of failure: /usr/local/lib/ruby/gems/3.4.0/gems/pups-1.4.0/lib/pups/exec_command.rb:138:in 'Pups::ExecCommand#spawn'
exec failed with the params {"tag" => "db", "cmd" => "if [ -f /root/install_postgres ]; then\n  /root/install_postgres && rm -f /root/install_postgres\nelif [ -e /shared/postgres_run/.s.PGSQL.5432 ]; then\n  socat /dev/null UNIX-CONNECT:/shared/postgres_run/.s.PGSQL.5432 || exit 0 && echo postgres already running stop container ; exit 1\nfi\n"}
bootstrap failed with exit code 1
** FAILED TO BOOTSTRAP ** please scroll up and look for earlier error messages, there may be more than one.
./discourse-doctor may help diagnose the problem.
064f80aa075fc97abf104570d60b7e79f936c8bd52efb55fffad3b293820ba82
```

tgxworld (approved):

PR #1108, opened 2026-08-11
DEV: Use Debian's Redis package instead of building our own

lightbulbjim:
Trixie contains Redis 8 so we upgrade at the same time.


fitzy101 (approved):
Does the Debian-provided `redis` package build with TLS enabled?

lightbulbjim:
TLS support is included but not enabled by default, same as our self-built version.

PR #1110, opened 2026-08-11
FIX: Ensure Redis log directory is created

lightbulbjim:
/var/log/redis needs to exist for certain tests in discourse/discourse to succeed. We were piggybacking off the Postgres install which discards the contents of /var/log via a tmpfs.


nattsw (approved):
I must be blind but I don't see where the directory is gettin gcreated

lightbulbjim:
The `--mount=type=tmpfs,target=/var/log` meant /var/log was mounted as a ramdisk and discarded after the command finished.

It would have been written like that because we don't care about Postgres cruft in /var/log.

PR #1116, opened 2026-08-19
PERF: Reduce compressed base image size by ~68MB (~5%)

davidtaylorhq:
- ~19MB - Mount tmpfs when installing nginx. Matches other apt steps in the docker file

- Same for postgres/redis, so that the 19MB doesn't move to their layer

- ~6MB - Mount tmpfs over /root/.cache when installing pups. `gem fetch` downloads the whole RubyGems spec index

- ~43MB - Set `install_extension_in_lib: false` in gemrc. By default RubyGems stores every compiled gem extension twice: once in extensions/ and again in the gem's own lib/. The extensions/ copy is the one that is actually used. This is planned to become the RubyGems 4 default: https://github.com/ruby/rubygems/issues/8921

- Pass `--package-import-method hardlink` to pnpm install. This does not change the size of our CI builds, which already hardlink. But it stops pnpm using reflinks on supported filesystems (btrfs, XFS, OrbStack). If reflinks are used, they are lost during the packaging process, so we silently double the cost of node_modules.

### DIFF
--- a/image/base/Dockerfile
+++ b/image/base/Dockerfile
@@ -128,7 +128,10 @@ RUN sed -i "s/^# $LANG/$LANG/" /etc/locale.gen &&\
     locale-gen
 
 # nginx
-RUN apt update &&\
+RUN --mount=type=tmpfs,target=/var/log \
+    --mount=type=tmpfs,target=/var/cache/apt \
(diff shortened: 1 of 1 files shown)


ZogStriP (approved):
🗜️

PR #1120, opened 2026-08-25
FIX: install PostgreSQL 15 server in postgres.15 template

Ethsim12:
`postgres.15.template.yml` currently assumes that the PostgreSQL 15 server
packages are installed by the base image.

Since the base image moved to PostgreSQL 18, selecting
`templates/postgres.15.template.yml` to postpone the PostgreSQL 18 upgrade
fails during bootstrap with:

    Errno::ENOENT: No such file or directory @ rb_sysopen -
    /etc/postgresql/15/main/postgresql.conf

Install the PostgreSQL 15 server packages in the PostgreSQL 15 template,
after removing the PostgreSQL 18 server packages.

This restores the intended PostgreSQL 15 holdback path.

### DIFF
--- a/templates/postgres.15.template.yml
+++ b/templates/postgres.15.template.yml
@@ -23,8 +23,9 @@ hooks:
 
 run:
   - exec: DEBIAN_FRONTEND=noninteractive apt-get purge -y postgresql-18 postgresql-contrib-18 postgresql-18-pgvector
-
+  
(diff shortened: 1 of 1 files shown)


Ethsim12:
One question: should this also purge `postgresql-client-18` and explicitly install `postgresql-client-15`, to match the approach in `postgres.13.template.yml`?

The current base already includes the PG15 client, so I’ve left the PR minimal for now, but I’m happy to align it with the older holdback template if that’s preferred.

lightbulbjim (approved):

PR #1124, opened 2026-09-01
update discourse/discourse builds

featheredtoast:
Remove docker manager plugin - this image is intended to only be updated via the command line.

Update to check if build passes at PR time

Build slim images for `discourse/discourse` - reduces size of image by ~50%


.github/workflows/build-web-only.yml
ducks, line 57: do we need to add `VERSION` to the env here as well?

ZogStriP (approved):

.github/workflows/build-web-only.yml
featheredtoast: oh, yes indeed 👍

ducks (approved):

PR #1125, opened 2026-09-04
DEV: run precompile:prettytext during build

featheredtoast:
Run asset:precompile:pretty_text alongside precompile:build during build time.

Ensures that the prebuilt prettytext cache is present in /tmp after build time.

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -249,7 +249,7 @@ run:
       tag: build
       hook: assets_precompile_build
       cmd:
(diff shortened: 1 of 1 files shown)


ZogStriP (approved):

corywright:
I'm getting an error when running `./launcher bootstrap app` after this change:
```
rake aborted!
Don't know how to build task 'assets:precompile:pretty_text' (See the list of available tasks with `rake --tasks`)
Did you mean?  assets:precompile
```
Reverting to a previous commit fixes the issue and the bootstrap successfully completes.

PR #1126, opened 2026-09-07
FIX: only run assets:precompile:pretty_text on latest

featheredtoast:
current esr does not have the assets:precompile:pretty_text job

### DIFF
--- a/templates/web.template.yml
+++ b/templates/web.template.yml
@@ -249,7 +249,14 @@ run:
       tag: build
       hook: assets_precompile_build
       cmd:
-        - su discourse -c 'SKIP_DB_AND_REDIS=1 bundle exec rake assets:precompile:build assets:precompile:pretty_text'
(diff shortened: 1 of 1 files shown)


ZogStriP (approved):

corywright:
I'm getting the following `fatal:` when using this latest commit:
```
I, [2026-09-07T15:35:58.429238 #1]  INFO -- : > cd /var/www/discourse && sudo -H -E -u discourse bash -c '
  if [[ $(git symbolic-ref --short HEAD) == stable ]] ; then
      SKIP_DB_AND_REDIS=1 bundle exec rake assets:precompile:build assets:precompile:pretty_text
  else
      SKIP_DB_AND_REDIS=1 bundle exec rake assets:precompile:build
  fi
'
fatal: ref HEAD is not a symbolic ref

